PR #4, opened 2023-12-11
Adding PR template

ksbeattie:
Adding the legal language to a PR template on the conditions to anyone contributions to this repo.


andrewlee94 (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+## Addresses Issue:
```
lbianchi-lbl, line 1: I this this should be: `## Resolves: ` in order to trigger the "closing keywords" functionality: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

ksbeattie: I actually changed it (from the idaes-pse example I started with) here to avoid it automatically closing issues, as it seems more often than not PRs only address parts of existing issues rather than close them entirely.

It's entirely a judgement call on if that's actually true or not.  I only felt slightly that "Addresses" is better than "Resolves" IMO.

PR #5, opened 2023-12-12
Move files from prommis/workspace 4491cb

lbianchi-lbl:


dangunter:
Two things I would like to change, maybe in a future PR

1. Get subpackage/module naming to be pep8, esp. UKy_flowsheet -> uky, but generally lowercase and avoid '_' where possible
2. Put tests in separate subdir, like it is in IDAES, WaterTAP, etc.

andrewlee94:
@dangunter Regarding tests, we probably also want to improve the tests overall as most of them are currently examples (which might have an assert statement) rather than unit tests.

lbianchi-lbl:
> Two things I would like to change, maybe in a future PR
> 
>     1. Get subpackage/module naming to be pep8, esp. UKy_flowsheet -> uky, but generally lowercase and avoid '_' where possible
> 
>     2. Put tests in separate subdir, like it is in IDAES, WaterTAP, etc.

@dangunter These have been addressed by 507d53d (2) and 017d66d (1).

MarcusHolly:
I don't think the filter press folder should be merged in. I think it was somehow merged accidentally into the old workspace repo, but it shouldn't be included until https://github.com/prommis/workspace/pull/5 is finished and approved. That PR is still a WIP, and I haven't touched it in a while. It's not really in "presentable condition" yet.

bpaul4:
Thank you for migrating everything over. I'd request that you remove the file "ree_test_flowsheet.py", and then the rest of the costing folder is good for the initial version. I will add a smoke test for the costing in a separate PR with a few checks, and then expand the tests next month to cover the entire costing code.

src/prommis/uky/autoscaling.py
```diff
@@ -0,0 +1,193 @@
+from pyomo.environ import (
```
MarcusHolly, line 1: @andrewlee94 Is it fine to merge the autoscaling file in this repo? I think at the very least, we could move it out of the uky folder since its meant for general use, not just this one application. Also, was this autoscaling tool intended to be in the IDAES repo eventually (or maybe it's already there and I just need to update the import in the uky flowsheet)?

ksbeattie:
I'd like to mention here that if anyone who has contributed code to the workspace repo that wants to preserve their contribution history (meaning keep their GitHub "grid of green") don't hesitate to speak up.  We should be able to work something out where you create your own PR separate from this one.

src/prommis/uky/autoscaling.py
andrewlee94: The autoscaling probably should not make it in, as it will eventually end up in IDAES as a utility tool.

MarcusHolly: The problem with not merging it in is that the flowsheet will almost definitely fail without it.

andrewlee94: Could you grab the Suffix it creates and just add that manually?

MarcusHolly: Not sure what you mean by that or how to do it

andrewlee94: You can display the Suffix using `model.scaling_factor.display()`. All the scaling factors are stored in that one Suffix (basically a dict, where the keys are components and ht value the scaling factor).

MarcusHolly: The flowsheet actually solves the same if I just remove the autoscaling, so I think my manual scaling may have already been very similar to the autoscaling

MarcusHolly: @lbianchi-lbl How would you like to handle this? Basically the following changes should be made:

1.) The autoscaling file should be removed

2.) The autoscaling import should be removed from the uky_flowsheet.py

3.) The solve function in the uky_flowsheet.py should look like this
`def solve(m):
    solver = SolverFactory("ipopt")
    results = solver.solve(m, tee=True)

    return results`

lbianchi-lbl: Thanks @MarcusHolly. I'm not yet sure if I'll do this as part of the Git history migration happening momentarily, or immediately afterwards. I've created #7 to track this.

Issue #6, opened 2023-12-12
Initial test failures from `prommis/workspace`


lbianchi-lbl:
![image](https://github.com/prommis/prommis/assets/48035537/703b8b41-ad8c-41d0-8327-93e9c45f39a2)

PR #14, opened 2023-12-14
Resolve Test Failures

MarcusHolly:
## Addresses Issue: 
#6 

- Resolves test faliures

### Legal Acknowledgement

   employer.


MarcusHolly:
Is it fine to add pytest.ini here? It's the only way I know of running an individual test file without running the whole repo. If there's another way and this file is actually not necessary, then please let me know. I've just copied the WaterTAP version of pytest.ini and changed the test path from watertap to prommis.

andrewlee94:
@lbianchi-lbl is the one to ask about `pytest.ini` files. I am actually surprised we did not have one already from when he set up the CI.

andrewlee94 (approved):

lbianchi-lbl:
> Is it fine to add pytest.ini here? It's the only way I know of running an individual test file without running the whole repo. If there's another way and this file is actually not necessary, then please let me know. I've just copied the WaterTAP version of pytest.ini and changed the test path from watertap to prommis.

@MarcusHolly that's a good question and (as many things) largely a matter of individual preference. Personally I'd recommend not using `pytest.ini`, for two main reasons:

1. The same exact functionality is provided by `pyproject.toml` (under the `tool.pytest.ini_options: https://github.com/prommis/prommis/blob/71253c5c84cfcc2bd87929f84ead00cb0b3ba73f/pyproject.toml#L77-L82
2. It introduces differences between test runs from a Git clone and a `site-packages` installation, which is the opposite of what we're trying to achieve (i.e. the ability to run `pytest --pyargs mypackage` regardless of how the `mypackage` package was installed)

Again, for the same goal of consistency between test runs (many things in pytest depend on the root directory and/or the current working directory, and we'd like to minimize implicit behaviors depending on either), the recommended way for running individual test files is to use the `-k` keyword, e.g. `pytest -k test_my_file.py` (this is also in the README.md in this repo).

pytest.ini
lbianchi-lbl, line 1: For the reasons detailed above, I suggest to delete `pytest.ini` before merging.

lbianchi-lbl (changes requested):
Everything looks good. Only thing is removing the `pytest.ini` file.

pytest.ini
MarcusHolly: Thanks for the explanation Ludovico! I've tested `pytest -k`, and it works fine.

lbianchi-lbl:
I forgot to add: before merging, we should ensure that all `@pytest.mark.known_issue` markers have been removed (these cause the marked tests to `xfail`, but if the tests now are expected to pass, they aren't needed anymore).

MarcusHolly:
@lbianchi-lbl One minor issue I'm seeing is that the pytest marks don't seem to be registered. I imagine this involves just adding a few lines to `pyproject.toml`?
image

lbianchi-lbl:
> @lbianchi-lbl One minor issue I'm seeing is that the pytest marks don't seem to be registered. I imagine this involves just adding a few lines to `pyproject.toml`? image

@MarcusHolly Good point. For the same reason of consistent behavior, the markers are registered in `src/prommis/conftest.py` (since that file will always be available, no matter the installation method): https://github.com/prommis/prommis/blob/71253c5c84cfcc2bd87929f84ead00cb0b3ba73f/src/prommis/conftest.py#L4-L6

You should be able to add the two markers following the existing key-value pairs (`marker_name: description`).

lbianchi-lbl (approved):
A sight for sore eyes... green, beautiful green, everywhere!

image

Issue #25, opened 2024-01-03
Add WaterTAP As Required Dependency for PrOMMiS


bpaul4:
Going forwards, a good amount of process modeling and cost estimation work will leverage WaterTAP and we should consider how to integrate the package to work with the PrOMMiS code.

Specifically, proposed models and tests introduced in #22 and #23 require WaterTAP as part of the CI suite.

lbianchi-lbl:
- There might be some complications if there are conflicts between the `idaes-pse` requirements between WaterTAP and this repository
- My plan would be to open a separate PR to add WaterTAP as a (dev-only? optional?) requirement, and then merge the changes back into #22 and #23

dangunter:
I also need this for the UI work since the Python UI export code is (at least for now) in the main watertap repo.

bpaul4:
Completed via #27.

PR #26, opened 2024-01-08
Interface to UI for UKy

dangunter:
## Summary/Motivation:
Shows the feasibility of using the WaterTAP UI for PROMMIS models.

- Add UI module, diagram, and some CSV exports in the `prommis.uky` subpackage
- At runtime, load this information into the WaterTAP UI

## Dependencies

https://github.com/watertap-org/watertap/pull/1292

## Legal Acknowledgement

   employer.


dangunter:
Currently waiting for #25 to resolve.

src/prommis/uky/tests/test_uky_flowsheet_ui.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from prommis.uky import uky_flowsheet_ui as ui
```
lbianchi-lbl, line 2:
```suggestion
import pytest
ui = pytest.importorskip("prommis.uky.uky_flowsheet_ui", reason="Flowsheet UI components not available")
```

lbianchi-lbl (changes requested):
Needs `pytest.importorskip` so that the tests pass when WaterTAP (which we're treating as an optional dependency) is not installed.

dangunter:
This PR seems to be stuck due to a codecov error that has these characteristics:
* Happens every time
* Only occurs with the Linux / Python 3.11 / Dev workflow

No idea what to do next, except force merge?
Thoughts, @lbianchi-lbl?

dangunter:
OK, it works now. We should approve it quick before it breaks again.

bpaul4 (commented):
@dangunter the code all works and the new content looks good, but I'm not sure how to generate the PNG file (or should it open a browser tab?). It would also be helpful if you could add some light documentation.

MarcusHolly (approved):
LGTM

dangunter:
@bpaul4 good point about the documentation, I will work on that.

I am not sure I understand the PNG question. Perhaps you are assuming it is autogenerated; it is just a picture. It should display in the initial "input" screen when that model is loaded into the WaterTAP UI.

bpaul4 (approved):
I'll trust that this works as intended, given that the tests are passing and everything looks good.

PR #27, opened 2024-01-08
Add WaterTAP dependency

lbianchi-lbl:
- Add WaterTAP dependency as `watertap` `extras_require` target (included by default in `requirements-dev.txt`)
- Change `idaes-pse` dev requirement to 2.3.0 (from `IDAES/idaes-pse@main`) to avoid conflicts with WaterTAP's own `idaes-pse` requirement

### Legal Acknowledgement

   employer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,11 @@ dependencies = [
     "idaes-pse >= 2.3.0rc1",
     "pyomo >= 6.7.0",
 ]
+[project.optional-dependencies]
+watertap = [
+    # TODO: update once 0.11.0 is available
+    "watertap==0.11.0rc1",
+]
 
 [tool.setuptools_scm]
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,4 +8,4 @@ nbmake==1.4.6
 
 idaes-pse @ git+https://github.com/IDAES/idaes-pse@main
 
---editable .
+--editable .[watertap]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,6 +6,6 @@ jupyter-book==0.15.*
 isort
 nbmake==1.4.6
 
-idaes-pse @ git+https://github.com/IDAES/idaes-pse@main
+idaes-pse==2.3.0
 
 --editable .[watertap]


dangunter:
Looks good to me but I'll add @MarcusHolly to also review the expectation that Prommis in main requires the latest release of idaes, not what's currently in idaes/main, to run. I tend to agree that will be less headaches all around, but it should be well-advertised since it's a change from what's there now.

lbianchi-lbl:
> Looks good to me but I'll add @MarcusHolly to also review the expectation that Prommis in main requires the latest release of idaes, not what's currently in idaes/main, to run. I tend to agree that will be less headaches all around, but it should be well-advertised since it's a change from what's there now.

Sounds good. On the topic of the IDAES requirement, one possible policy/heuristic could be:

>Assume the most recent stable release, unless tests start breaking, in which case we'd switch to `IDAES/idaes-pse@main` (or our own `prommis/idaes-pse` fork).

I'd also note that, as of WaterTAP 0.11.0rc1, we do need to use `idaes-pse==2.3.0` (or, more precisely, any 2.3.x release) in order to satisfy the `idaes-pse` requirement in `watertap`. We can (and probably should) discuss the possibility of relaxing this on the WaterTAP side, but that requires a bit of thought since it inevitably involves some tradeoffs for WaterTAP's other dependent projects.

MarcusHolly:
I don't see any issue with using `idaes-pse==2.3.0`. As far as using the most recent stable release and reverting back only if tests fail, it seems like the ideal "solution", but I'll admit that I'm not sure how this would actually look in practice. Would we expect to run into failures/conflicts on a semi-regular basis? I'm assuming it would happen very infrequently, so it should be fine.

lbianchi-lbl:
> I don't see any issue with using `idaes-pse==2.3.0`. As far as using the most recent stable release and reverting back only if tests fail, it seems like the ideal "solution", but I'll admit that I'm not sure how this would actually look in practice. Would we expect to run into failures/conflicts on a semi-regular basis? I'm assuming it would happen very infrequently, so it should be fine.

The case I had in mind was e.g. PrOMMiS adding code (e.g. `prommis.model_A`) that uses an IDAES feature  (e.g. `idaes.feature_A`) that has been merged into `IDAES/idaes-pse:main` before that feature makes its way into an IDAES release, let's say between IDAES 2.3 and 2.4.

- In this scenario, PrOMMiS will cease to be compatible with the latest IDAES release, and we expect CI runs to start failing as soon as `prommis.model_A` is added to the PrOMMiS codebase
- At that point, to restore compatibility (i.e. for CI checks to start passing again) we'll need to switch from requiring `idaes-pse==2.3.*` to `idaes-pse @ https://github.com/IDAES/idaes-pse@main`

My point was that we should make sure that our test suite/CI checks are written in such a way that the compatibility is actually enforced, which most of the time just amounts to "write tests when adding new features" (i.e. nothing special other than what we're generally doing).

In the simplistic example above:

- If there's no test for `prommis.model_A`, the test suite would "fail to fail" (i.e. pass even when it was expected to fail) regardless of whether `idaes.feature_A` is available or not
- This would be the scenario where our test suite/CI checks is not representative of the compatibility between PrOMMiS and IDAES that I was referring to, and that we should do our best to avoid

**TL;DR** Let's make sure we write enough tests, so that our CI can be relied upon to answer the question "does PrOMMiS work as expected?"

MarcusHolly:
Alright, thanks for the thorough explanation. I'll be more strict with future PRs to make sure they have adequate testing.

lbianchi-lbl:
@dangunter @MarcusHolly @ksbeattie @bpaul4 Any chance you could re-review? It looks like this could be merged, but I wanted to confirm with you first.

MarcusHolly (approved):

dangunter (approved):
LGTM

PR #30, opened 2024-01-16
Dynamic modification

ArkoprabhoDasgupta:
## Addresses Issue: 
In this branch, I have added the dynamic aspect of the solvent extraction model. There are two problems regarding the dynamic aspect running properly.
1. The material accumulation term needs a function called get_material_density_term in the property package. This function is not present in the leaching property package.
2. There is a unit issue in the model for solvent extraction model which is coming from the mscontactor model. The unit of volume is meter cubed, whereas all concentration units in the REESim excel file is as g/L or Kg/L.
Please kindly suggest what should be done to fix these errors.

-
-

### Legal Acknowledgement

4. I represent I am authorized to make the contributions and grant the license. If my employer has
   employer.

### DIFF
--- a/src/prommis/solvent_extraction/sx_model_flowsheet.py
+++ b/src/prommis/solvent_extraction/sx_model_flowsheet.py
@@ -1,4 +1,4 @@
-from pyomo.environ import ConcreteModel, SolverFactory
+from pyomo.environ import ConcreteModel, SolverFactory, units, TransformationFactory
 
 from idaes.core import FlowDirection, FlowsheetBlock
 from idaes.core.initialization import InitializationStatus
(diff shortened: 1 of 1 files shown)


ArkoprabhoDasgupta:
> @ArkoprabhoDasgupta I don't have any experience with dynamic modeling myself, but I can try to address your questions since Andrew is still out.
> 
> 1.) It should be fairly simple to just add a `get_material_density_terms` into the leaching property package. For an example of what this might look like, see https://github.com/watertap-org/watertap/blob/main/watertap/property_models/activated_sludge/modified_asm2d_properties.py
> 
> 2.) Do you have the exact error message? Regardless, two possible solutions would be to either convert the units after importing them from REESim (using `units.convert` or `pyunits.convert` depending on how you have the imports set up) or directly modifying the Excel itself (like creating a new tab where you convert the units to m3 and importing from this new tab).
> 
> Let me know if these "solutions" seem feasible.



> @ArkoprabhoDasgupta I don't have any experience with dynamic modeling myself, but I can try to address your questions since Andrew is still out.
> 
> 1.) It should be fairly simple to just add a `get_material_density_terms` into the leaching property package. For an example of what this might look like, see https://github.com/watertap-org/watertap/blob/main/watertap/property_models/activated_sludge/modified_asm2d_properties.py
> 
Yes adding this function will solve the problem. But this property package is a part of another unit operation. Can I add this term for now? I am wondering that whether it will upset any other unit operations being performed where this package is involved.

> 2.) Do you have the exact error message? Regardless, two possible solutions would be to either convert the units after importing them from REESim (using `units.convert` or `pyunits.convert` depending on how you have the imports set up) or directly modifying the Excel itself (like creating a new tab where you convert the units to m3 and importing from this new tab).
> 
> Let me know if these "solutions" seem feasible.

I will show you the exact error message in a snapshot in Slack. Volume is a term which I am fixing in the flowsheet as an input. I do not know whether I have a scope of doing units.convert in it. The place where I think it can be given is in the main model, that is the mscontactor one.

Issue #13, opened 2023-12-14
UKy Flowsheet Refinements


MarcusHolly:
- [x] Update roaster model to accept a liquid/moisture inlet @jinliangma #88 
- [x] Add `.report()` functionality to the roasting model @jinliangma #88 
- [x] Fix scaling so that no warnings are displayed when running the flowsheet #47 
- [x] Add pH adjustment blocks (mixers adding an acid or a base stream) #47 
- [x] Image should be consistent when showing mixers (we should show all) #86 
- [x] Add name of the mixers as all the other unit models #86 
- [x] Make stream naming a little more consistent or easier to follow (maybe add an extra term in name to show what part of the flowsheet we are on) #86 
- [x] Add comment on why we added eps term #86 
- [x] In the notebook we could add an explanation on the initialization loops #86

PR #34, opened 2024-02-05
UKy Flowsheet Jupyter Notebook

MarcusHolly:
## Summary/Motivation:
Creates Jupyter notebook documentation that walks through how to set up and run the UKy flowsheet

### Legal Acknowledgement

   employer.


MarcusHolly:
Current failures are unrelated to these changes

bpaul4:
The failures are due to recent updates made in the Pyomo 6.7.1 release which replace implicit sets created by `IndexedComponent` objects with anonymous sets. One test fails because the `_index` object doesn't exist anymore, the other fails because it expects a `RuntimeError` that is no longer thrown by Pyomo.

I suggest adding a skip flag for those two tests to resolve this PR, and I'll create an issue for them.

andrewlee94:
@MarcusHolly @bpaul4 Whilst the test failures here might not be related to this PR, they do reveal some significant issues in the code that is failing. Most obviously, I see a lot of warnings about implicit replacement of Pyomo components that needs to be corrected wherever it is coming from. This probably needs to be broken out as a new issue and addressed fairly quickly.

MarcusHolly:
So should I skip those tests for now as Brandon suggested or just wait for someone (probably Brandon) to merge the PR fixing these issues?

bpaul4:
@andrewlee94 there is a test that purposefully calls costing on the same model twice, and before the 6.7.1 update it would fail when it tries to create an `_index` object that already exists. Now, it doesn't fail but it's still bad practice, we could add our own error message instead.

It looks like the warnings are from the test that costs the model twice, so it's not a bug with the code itself.

I opened an issue: #40

andrewlee94:
@bpaul4 What exactly were you trying to test by doing that, and what was preventing this from occurring in the past? You should not be relying on Pyomo's implicit replacement warnings for this, as I would consider them unreliable.

@MarcusHolly We should generally not skip tests wherever possible, as it is easy to forget to re-enable them. Unless there is something time critical and @bpaul4 cannot get a fix done in time, I would say we should wait for now.

lbianchi-lbl:
@MarcusHolly FYI, there's a couple of steps needed for the notebook being added in this PR to be built and tested as part of the documentation:

- Move the `uky_flowsheet.ipynb` file to the `docs/tutorials/` subdirectory
- Add an entry for it in the `docs/_toc.yml` file, e.g. 
  ```diff
  format: jb-book
  root: index
  chapters:
    - file: tutorials/diafiltration
    - file: tutorials/diafiltration-solution
  +  - file: tutorials/uky_flowsheet
  ```

bpaul4:
@andrewlee94 I will put up a fix ASAP.

lbianchi-lbl:
@MarcusHolly a couple of small changes for the notebook:

- Remove the trailing `:` in the title
- Rename `docs/tutorials/uky_flowsheet.ipynb` to `docs/tutorials/uky_flowsheet-solution.ipynb`
  - Because of how the notebook testing is set up for tutorials, notebooks that don't contain the suffix `-solution` are assumed to be exercises (i.e. they're not expected to run before being filled in by users)
  - We can (and IMO should) revisit this assumption soon, but for the purposes of getting this PR tested and merged, I suggest renaming the notebook so that it matches the current convention/requirement

bpaul4 (commented):
Overall it looks very nice and the code is extremely detailed. I have a few comments:

1. Adding more descriptive text before each code block, would be very helpful in explaining what is happening in each step.
2. There are a lot of warnings that are printed after the solver call.
3. Does a report() method exist for these units? It would look cleaner than the display() call.

MarcusHolly:
> Overall it looks very nice and the code is extremely detailed. I have a few comments:
> 
> 1. Adding more descriptive text before each code block, would be very helpful in explaining what is happening in each step.
> 2. There are a lot of warnings that are printed after the solver call.
> 3. Does a report() method exist for these units? It would look cleaner than the display() call.

1.) I can definitely expand on this
2.) Yea... I'm going to have to find a way to clean that up eventually. The issue is that a lot of concentrations variables are outside of their lower bounds (zero) because they are going to a really small, negative value. Is there a quick fix for this @andrewlee94 
3.) I don't think the report functionality currently exists - I run into the following: `idaes.core.util.exceptions.ConfigurationError: Unit model fs.roaster does not have the standard Port names (inlet and outlet). Please contact the unit model developer to develop a unit specific stream table.`

andrewlee94:
@MarcusHolly 

Regarding 2, that is a sign you need to scale the problem.
Regarding 3, you should reach out to Jinliang and get him to add a `report` method for the roaster. In the mean time, you will just have to do it by hand.

MarcusHolly:
For 1, I added a bit more detail, but let me know @bpaul4 if there's a specific section you think needs more explaining.

For 2, I don't think I have the time to fully resolve this now. Scaling is already a bit messy as it is, but I've made note of this issue in #13 

For 3, @jinliangma Can you add the report functionality to the roaster model? I've also added this to #13  @andrewlee94 Can I just leave the `m.fs.roaster.display()` in the mean time?

andrewlee94:
@MarcusHolly If it is commented out, with a note listing the reason it is probably OK to leave the line there as a reminder for the future.

agarciadiego (approved):

lbianchi-lbl (changes requested):
@MarcusHolly It looks like the image is not showing up (at least in the ReadTheDocs preview built from this PR):

image

MarcusHolly:
> @MarcusHolly It looks like the image is not showing up (at least in the ReadTheDocs preview built from this PR):
> 
> image

Thanks for pointing this out. It's working on my end now.

lbianchi-lbl (approved):
Thanks @MarcusHolly, that seems to have fixed it!

image

PR #35, opened 2024-02-05
Add code coverage infrastructure

lbianchi-lbl:
- Add `codecov.yml` configuration file
- Modify CI workflow to add coverage report generation and upload
- Add `pytest-cov` dev dependency

### Legal Acknowledgement

   employer.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -114,6 +114,7 @@ jobs:
             runner-image: windows-2022
           - install-variant: dev
             python-version: '3.11'
+            os: linux
             coverage: true
     steps:
       - uses: prommis/prommis/.github/actions/install@main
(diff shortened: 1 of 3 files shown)


bpaul4 (approved):
I think this looks great, although I couldn't find a coverage report on the current branch (maybe because there is no prior report to compare to). If the CI is manually retriggered, would it compare against the prior run and show the coverage number?

lbianchi-lbl:
> I think this looks great, although I couldn't find a coverage report on the current branch (maybe because there is no prior report to compare to). If the CI is manually retriggered, would it compare against the prior run and show the coverage number?

Good point. I believe the comparison only happens if a coverage report is available for a commit on the base branch of a PR, which typically means the default branch (`main` in our case). So, to get things rolling, I think we just have to merge this PR in and then we should be able to see the comparison for the other PRs (a bit of a chicken-and-egg thing).

PR #36, opened 2024-02-06
Add Costing to UKy Flowsheet

MarcusHolly:
## Summary/Motivation:
Adds costing to the UKy flowsheet and changes the number of finite elements in leaching from 1 to 2.

### Legal Acknowledgement

   employer.


src/prommis/uky/uky_flowsheet.py
```diff
@@ -862,5 +874,782 @@ def display_results(m):
...
+            "source": 1,
+            "n_equip": 1,
+            "scale_down_parallel_equip": False,
+            "CE_index_year": CE_index_year,
+        },
+    )
+
+    # TODO: How should this be handled considering there are multiple feeds with varying concentrations?
+    m.fs.feed_input = Var(initialize=500, units=units.ton / units.hr)
+    m.fs.feed_grade = Var(initialize=356.64, units=units.ppm)
```
MarcusHolly, line 1436: Not sure how to handle this... Is this referring to the feed of the actual plant as opposed to the feed of this flowsheet (what's being fed into the leaching model)? If so, then I assume we can just leave this as is.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -862,5 +874,782 @@ def display_results(m):
...
+
+    # TODO: Replace this number with the sum of flowsheet.fs.roaster.flow_mol_comp_product (need to convert units)
+    m.fs.recovery_rate_per_year = Var(
+        initialize=39.3
+        * units.kg
+        / units.hr
+        * 0.8025  # TREO (total rare earth oxide), 80.25% REE in REO
+        * m.fs.annual_operating_hours,
+        units=units.kg / units.yr,
+    )
```
MarcusHolly, line 1457: This seems like the recovery rate of REEs from the plant as a whole. Should the 39.3 be replaced by the mass flow of the product stream from the roaster unit?
image

src/prommis/uky/uky_flowsheet.py
```diff
@@ -862,5 +874,782 @@ def display_results(m):
...
+    # TODO: Replace these with values from the flowsheet?
+    m.fs.solid_waste = Var(
+        m.fs.time, initialize=11136 / 24, units=units.ton / units.hr
+    )  # non-hazardous solid waste
+    m.fs.precipitate = Var(
+        m.fs.time, initialize=732 / 24, units=units.ton / units.hr
+    )  # non-hazardous precipitate
+    m.fs.dust_and_volatiles = Var(
+        m.fs.time, initialize=120 / 24, units=units.ton / units.hr
+    )  # dust and volatiles
```
MarcusHolly, line 1507: I imagine these should be replaced with values from the flowsheet. What counts as solid waste - the amount of solid leaving the purge streams (purge streams are currently 100% liquid)?  Is the precipitate the sum of the solids leaving from the filter press? Dust and volatiles will just be the `flow_mol_comp_dust` stream from the roaster (after converting its units)

src/prommis/uky/uky_flowsheet.py
```diff
@@ -862,5 +874,782 @@ def display_results(m):
...
+        mixed_product_output_rates=mixed_product_output_rates,
+        mixed_product_sale_price_realization_factor=0.65,  # 65% price realization for mixed products
+        # arguments related to total owners costs
+        land_cost=m.fs.land_cost,
+        resources=resources,
+        rates=rates,
+        # TODO: Verify this cost
+        prices={
+            "dummy": 1 * getattr(units, "USD_" + CE_index_year) / units.kg,
+        },
```
MarcusHolly, line 1594: Should probably rename this from "dummy", but do we know what this cost actually should be?

src/prommis/uky/uky_flowsheet.py
andrewlee94: Shouldn't these be tied to flowsheet variables too? At the moment it seems like the costing is entirely separate from the flowsheet.

andrewlee94 (changes requested):
This might just be my lack of understanding, but right now it seems like this costing is not coupled to the flowsheet it is modeling (i.e. it is a static costing for some nominal case). All the sizing and design variables in the costing need to be tied to the actual flowsheet variables.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Yea, I just carried over these values from the costing test file since I wasn't sure what value to use here

src/prommis/uky/uky_flowsheet.py
```diff
@@ -20,6 +20,9 @@
 from pyomo.environ import (
     ConcreteModel,
     Constraint,
+    check_optimal_termination,
```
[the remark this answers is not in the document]
MarcusHolly: Yea, I'll move over the check_optimal_termination to the test file later in this PR

MarcusHolly:
> This might just be my lack of understanding, but right now it seems like this costing is not coupled to the flowsheet it is modeling (i.e. it is a static costing for some nominal case). All the sizing and design variables in the costing need to be tied to the actual flowsheet variables.

You're understanding correctly. I've addressed a couple of these in my latest commit, but a lot of my earlier comments are asking about which flowsheet variable I should be using in the costing.

src/prommis/uky/uky_flowsheet.py
bpaul4: Yes, `m.fs.feed_input` refers to the total plant feed, which is different than the leaching feed when you have pre-roasting (some mass is lost). For this case I think it would be your flowsheet feed.

src/prommis/uky/uky_flowsheet.py
bpaul4: Solid waste would be anything that would end up in a landfill, so most of your bulk solid that is separated. Precipitate is solid produced from treating wastewater or liquid waste streams, i.e. removing dissolved metal that end up on filter cakes.. As you mentioned, the dust and volatiles are produced by the roasters. The solid waste is definitely the largest output stream - the case study found that the total process yield of feed to product streams gave 92.8% solid waste, 6.1% precipitate waste, 1.0% dust and volatiles waste, and <1% actual REE products.

src/prommis/uky/uky_flowsheet.py
bpaul4: You could leave this argument as None or remove it if the chemicals list is entirely components with built-in prices (see https://github.com/prommis/prommis/blob/1d05932488e348a572f315969c98abb9b70dd164/src/prommis/uky/costing/ree_plant_capcost.py#L1997). I'm currently working on support for passing the total chemical cost or total waste cost as expressions similar to the land cost, and plan to have a PR up for that shortly.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -862,5 +872,782 @@ def display_results(m):
...
+    m.fs.L_tank_mixers.power = Var(initialize=474, units=units.hp)
+    m.fs.L_tank_mixers.power.fix()
+    m.fs.L_tank_mixers.costing = UnitModelCostingBlock(
+        flowsheet_costing_block=m.fs.costing,
+        costing_method=QGESSCostingData.get_REE_costing,
+        costing_method_arguments={
+            "cost_accounts": L_tank_mixer_accounts,
+            "scaled_param": m.fs.L_tank_mixers.power,
+            "source": 1,
+            "n_equip": 3,
```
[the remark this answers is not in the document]
bpaul4: I appreciate the input @andrewlee94 and can include your suggestions as part of other cost modeling improvements that I'm working on. Feel free to reach out to me to discuss, thanks!

src/prommis/uky/uky_flowsheet.py
MarcusHolly: I don't believe there would be any solid waste in this flowsheet based on how things are currently set up. Solid only leaves the system via the roaster or the filter press as a filter cake. Here's what the breakdown looks like for where solids end up: 
image

andrewlee94: There is the waste solids from the leaching stage filter press.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,6 +871,733 @@ def solve(m):
...
+            "n_equip": flowsheet.fs.leach.config.number_of_finite_elements,
+            "scale_down_parallel_equip": False,
+            "CE_index_year": CE_index_year,
+        },
+    )
+
+    # 4.3 is UKy Leaching - Tank Mixer
+    L_tank_mixer_accounts = ["4.3"]
+    m.fs.L_tank_mixers = UnitModelBlock()
+    m.fs.L_tank_mixers.power = Var(initialize=474, units=units.hp)
```
MarcusHolly, line 921: I've updated hard-coded values where possible, but there are still a lot of values that we don't (to my knowledge) have values for in the flowsheet. For instance, what should this horsepower be? The hard-coded numbers seem to represent a system that is orders of magnitude larger than this system, so most of these values will likely need to be changed.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Yea, that is entirely what the precipitate mass flow is comprised of. I didn't count that as waste... I suppose we could count it as waste but then we would have no precipitate mass flow.

andrewlee94: We have two solid outlet streams. There is the final precipitate leaving the roaster and there is also the left-over solids from the leaching stage.

MarcusHolly: I think the issue here was that the costing package defined 3 distinct solid outlet streams (solid waste, precipitate, and dust & volatiles)

- Dust and volatiles = the dust stream from the roaster
- Solid waste = None???
- Precipitate = filter cake from leaching

I don't think the product REEs from the roaster are represented in any of these 3 streams, but feel free to correct me if I'm mistaken about something @bpaul4

MarcusHolly: For ease of reference, this is how they are currently defined. I suppose I could just get rid of the solid waste stream instead of setting it to zero tho.
image

bpaul4: I believe waste from the roaster can be considered dust & volatiles, waste from any leaching can be considered solid waste, and waste from any precipitation or extraction can be considered precipitates. The solid waste should be most of the incoming solid feed, in terms of mass flow, and the precipitate waste should be much smaller.

MarcusHolly: Ok, then the filter cake from the leaching section needs to be updated so that it's treated as solid waste. There is no waste from precipitation or solvent extraction, so the precipitate waste will become 0 (i.e. solid waste and precipitate will be flipped in the screenshot above)

andrewlee94: Yes, the solids from leaching is solids waste. The "precipitate stream" is a secondary solid waste coming from treatment of the purge stream of the combined residual solid in the leaching filter cake and the recycled stream from the SX rougher.

MarcusHolly:
@andrewlee94 @bpaul4 I'm not sure how we want to handle this PR:

1. Merge it in since the costing is functional even if all the costing parameters aren't necessarily accurate
2. Leave this PR up until we can identify more accurate costing parameters for the unit models.

I lean more towards 1, although maybe this warrants making an issue just so that this doesn't fall through the cracks.

andrewlee94:
@MarcusHolly If it does get merged in the current state it definitely needs an issue. How much more needs to be done, and what are the holdups?

MarcusHolly:
The holdup is just me not knowing what the parameter values should be. I could dig through the reports and try to pull out these values, but I'm assuming that's where Brandon got the original values from, which aren't even the right magnitude in most cases. For example the volume of our leaching PE tanks in this flowsheet was 100gal, but the default value was 164805gal.

I've tried to link these parameter values to values directly from the flowsheet, but that only covers ~10% of the parameters (I may have missed a few)

andrewlee94:
What variables are we talking about here? A lot of the ones I saw in my last review were design variables that should be linked to the existing flowsheet variables (so that when we do the optimization the unit design is coupled to the cost).

MarcusHolly:
Power of the tank mixer as shown here: https://github.com/prommis/prommis/pull/36#discussion_r1481767631 or even some "simple" values like the volume of the solvent extraction cleaner aren't specified in the flowsheet (or the unit model files) from what I can tell.

andrewlee94:
For things like those you should first reach out to the model developer - they should be the ones to provide that information if possible.

MarcusHolly:
Many of the costing parameters used in this PR still need to be re-evaluated/updated when additional data is available. I've tried to use the values from the flowsheet itself where possible and have reduced the remaining parameters by 1-3 orders of magnitude as an extremely rough guess of what those values should be on a pilot scale. Let me know if there are any other changes I should make before merging in this preliminary costing.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -862,5 +874,719 @@ def display_results(m):
...
+        m.fs.solid_waste,
+        m.fs.precipitate,
+        m.fs.dust_and_volatiles,
+        m.fs.power,
+    ]
+
+    # define product flowrates
+    # TODO: Need to revisit this after getting costs for products defined in mw_comp_product in ree_feed_roaster
+
+    pure_product_output_rates = {
```
bpaul4, line 1472: @MarcusHolly if we're scaling down the inputs, we should also scale down these output rates. If the flowsheet provides these values, that would be even better.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -862,5 +874,719 @@ def display_results(m):
...
+            "nonhazardous_precipitate_waste",
+            "dust_and_volatiles",
+        ],
+        recovery_rate_per_year=m.fs.recovery_rate_per_year,
+        CE_index_year=CE_index_year,
+    )
+
+    # define reagent fill costs as an other plant cost so framework adds this to TPC calculation
+    m.fs.costing.other_plant_costs.unfix()
+    m.fs.costing.other_plant_costs_rule = Constraint(
```
bpaul4, line 1548: These would scale down or import from the flowsheet as well, they are the cost of pre-filling these unit operations with chemicals prior to start-up. If the chemicals exist in the dictionary [here](https://github.com/prommis/prommis/blob/b903c34908f76c6dd6936a13d41baa64f45a12fd/src/prommis/uky/costing/ree_plant_capcost.py#L2048), you can use those prices to estimate the absolute costs and update these values.

bpaul4 (commented):
A couple comments on down-scaling some costing inputs, if you plan on doing that before merging this in.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Some of the default sale prices for oxide products are missing - not sure if we want to resolve that now or later

bpaul4: Which products are missing from the [sale price dictionary](https://github.com/prommis/prommis/blob/b903c34908f76c6dd6936a13d41baa64f45a12fd/src/prommis/uky/costing/ree_plant_capcost.py#L1680)?

MarcusHolly: Al2O3, Ce2O3, Fe2O3, & Pr2O3

andrewlee94: I am not sure if we want to assign prices to Al and Fe - those are impurities we want to avoid, and thus really represent a cost penalty.

MarcusHolly: Are Ce2O3 and Pr2O3 meaningfully different from CeO2 and Pr6O11? We do have sale prices for the latter oxides, but I assumed they weren't applicable.

andrewlee94: That depends on what the stable oxides of each species are. I would suspect that if we have sale prices for something it is probably the stable oxide.

MarcusHolly: CeO2 and Pr6O11 are the stable oxides, so I'll just use the pricing for those

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1467,31 +1467,193 @@ def add_costing(flowsheet):
...
-    # TODO: Need to revisit this after getting costs for products defined in mw_comp_product in ree_feed_roaster
+    # TODO: Default sale prices for some components in flow_mol_comp_product are missing
+
+    ce_product = value(
+        units.convert(
+            flowsheet.fs.roaster.flow_mol_comp_product[0, "Ce"]
+            * 328.24
+            * units.g
+            / units.mol,
+            to_units=units.kg / units.hr,
```
andrewlee94, line 1478: I have a few questions/comments here.

1. Why are you getting the value of the flowrate here? This means that this is a static value and is no longer coupled to the flowsheet.
2. Looking at the next section, you then take these static values and multiply by the units. You can get this directly by using the `as_quantity(var)` function instead of `value()` - this returns a `pint` `Quantity` which has both the magnitude and units.
3. Finally, I would be inclined to name these `Ce_product` as the capitalization is part of the name. This is mostly useful for things like `Y_product` which could be a bit more ambiguous.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1467,31 +1467,193 @@ def add_costing(flowsheet):
...
-        "Eu2O3": 0.00199 * units.kg / units.hr,
-        "Gd2O3": 0.00000 * units.kg / units.hr,
-        "Tb4O7": 0.00801 * units.kg / units.hr,
-        "Dy2O3": 0.00000 * units.kg / units.hr,
-        "Ho2O3": 0.00000 * units.kg / units.hr,
-        "Er2O3": 0.00000 * units.kg / units.hr,
-        "Tm2O3": 0.00130 * units.kg / units.hr,
-        "Yb2O3": 0.00373 * units.kg / units.hr,
-        "Lu2O3": 0.00105 * units.kg / units.hr,
+        "CeO2": ce_dust * units.kg / units.hr,
```
andrewlee94, line 1648: Looking at how this is used later, this appears to be wrong to me. Why are you passing static values to the costing, and not the actual variables?

MarcusHolly: What's the issue with using static values instead of the variables? Shouldn't the two be equivalent at steady state?

andrewlee94: No - the static value is only valid at the time you access it. Most importantly, it means if you try to do an optimization these values will not change; i.e. you can change the actual flowsheet performance but the costs will not change to reflect this.

src/prommis/uky/uky_flowsheet.py
andrewlee94: It should be able to do this, however I would not do it that way myself. `as_quantity` returns a `Quantity` so it us the units in it anyway so you can do the conversion after the fact.

However, this ties into the question overall of whether you should be using `value` or `as_quantity` at all.

andrewlee94: Also, can you show the rest of the traceback? That last line is coming from `pint` alone, so it could be a `pint` issue.

MarcusHolly: I think we could just get away with not using `value `or `as_quantity`, but it's currently running into a DOF issue when I do this, so I need to figure out why.

lbianchi-lbl:
@MarcusHolly do you think this will make the March release? It'd need to get merged within the next week for this.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,956 @@ def display_results(m):
...
+        },
+    )
+
+    # 3.2 is UKy Roasting - Conveyors
+    R_conveyors_accounts = ["3.2"]
+    m.fs.R_conveyors = UnitModelBlock()
+
+    flow_3_2 = value(
+        units.convert(
+            flowsheet.fs.roaster.flow_mas_product[0]
```
bpaul4, line 1280: There appears to be a typo in the two roaster flowrates, and I'm surprised that this doesn't fail the tests.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,956 @@ def display_results(m):
...
+        },
+    )
+
+    feed_input = units.convert(
+        flowsheet.fs.leach_solid_feed.flow_mass[0]
+        + flowsheet.fs.leach_liquid_feed.flow_vol[0]
+        * flowsheet.fs.leach.liquid[0, 1].dens_mol,
+        to_units=units.ton / units.hr,
+    )
+    # Feed grade is grams of solute per 1000 grams of solution (solution ~ water since dilute concentrations)
```
bpaul4, line 1382: The feed grade is the concentration of REE in the feedstock ore. This flowsheet has no Crushing & Screening or Pre-Roasting operations, so it would be based on the solid flow into the leaching unit. Basically, we're assuming Crushing & Screening and Pre-Roasting have no losses since we don't have those units here.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,956 @@ def display_results(m):
...
+
+    # for convenience
+    m.fs.annual_operating_hours = Param(
+        initialize=hours_per_shift * shifts_per_day * operating_days_per_year,
+        mutable=True,
+        units=units.hours / units.a,
+    )
+
+    recovery_rate = units.convert(
+        flowsheet.fs.roaster.flow_mas_product[0], to_units=units.kg / units.hr
```
bpaul4, line 1408: Same as before, is this a typo and should this have caused a test to fail?

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,956 @@ def display_results(m):
...
+        m.fs.time, initialize=solid_waste, units=units.ton / units.hr
+    )  # non-hazardous solid waste
+
+    m.fs.precipitate = Var(
+        m.fs.time, initialize=0, units=units.ton / units.hr
+    )  # non-hazardous precipitate
+
+    dust = value(
+        units.convert(
+            flowsheet.fs.roaster.flow_mas_dust[0], to_units=units.ton / units.hr
```
bpaul4, line 1446: Same as roaster comments above.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,956 @@ def display_results(m):
...
+        "Nd2O3": m.fs.Nd_product,
+        "Pr6O11": m.fs.Pr_product,
+        "Sm2O3": m.fs.Sm_product,
+        "Gd2O3": m.fs.Gd_product,
+        "Dy2O3": m.fs.Dy_product,
+    }
+
+    m.fs.Ce_dust = Param(
+        default=units.convert(
+            flowsheet.fs.roaster.flow_mol_comp_dust[0, "Ce"]
```
bpaul4, line 1602: Just want to verify that the "dust" streams are supposed to be product streams and not volatile/dust waste. The UKy report define pure products as targeted oxides that should be separated into pure streams, and mixed products as other oxides that have notable value but may be solid as a mixed basket or are not worth separating.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: This is the correct name of the unit property, but the unit property itself has this typo. I'll try to fix it now - it shouldn't take long.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: So just to be clear, feed grade would only be the concentration of REEs, meaning it would exclude the concentrations of the contaminants (Al, Fe, and Ca)? And would it be the concentration of the REE or the REE oxide?

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Same as the roaster - the property itself has this typo.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: So are you saying the dust streams should be mixed products (the current implementation) or pure product?

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Fixed.

src/prommis/uky/uky_flowsheet.py
bpaul4: Yes, feed grade would only be the concentration of REEs in the original bulk solid, where contaminants, solid waste, ash, etc are part of the bulk solid but only saleable products are considered as REEs. And yes, REE not REE oxide.

src/prommis/uky/uky_flowsheet.py
bpaul4: The roaster produces a waste dust and also a sintered product. It seems from https://github.com/prommis/prommis/blob/main/src/prommis/roasting/ree_oxalate_roaster.py#L41 that `flow_mol_comp_dust` is a waste stream and part of the dust & volatiles waste flow. The solid product contains both pure and mixed products, which are separated downstream into pure REO streams and a mixed basket of everything else. I don't think we have the downstream units here yet, so you could consider assigning a few specific components as "pure"/"desired" products and everything else as "mixed" for now.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Maybe I'm just misunderstanding, but isn't ppm a weird/incorrect unit of measurement to be using for a solid feed stream (i.e. we can't get the REE concentration in mg/mL for the solid feed, but we could get something like kg/hr). So I'm a bit confused what the units of feed grade should be here.

src/prommis/uky/uky_flowsheet.py
andrewlee94: The roaster does not really produce both mixer and pure products - it just produces a mixer product which can be further separated later (it is not like there are two separate phases of material). For the UKy process, the only product is mixer REEs (which is the desired product); there is no "pure" product of interest here.

Separation of the mixer REEs into more targeted grouping required an entire additional process that we are not modeling here (and would often happen at a different geographic location).

src/prommis/uky/uky_flowsheet.py
andrewlee94: No - remember that ppm is an application specific unit. Here it means micrograms per gram and not mg/L.

src/prommis/uky/uky_flowsheet.py
bpaul4: @andrewlee94 thank you for the explanation. @MarcusHolly you can assign all products as mixed products then. Set the argument `pure_product_output_rates` as an empty dictionary, or a single component with zero flowrate. I made the pure and mixed product rates required arguments and an error is thrown if one of them is not set so that users know this is something that should be considered.

MarcusHolly: I just set it as an empty dictionary.

Agfritz (commented):
Hi Marcus,
I have reviewed the PR and added comments. Most were minor and may also be the result of my limited experience with IDAES/ PrOMMiS coding conventions. Happy to discuss further if they are unclear!
Best,
Alison

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,839 @@ def display_results(m):
...
+            "n_equip": 3,
+            "scale_down_parallel_equip": False,
+            "CE_index_year": CE_index_year,
+        },
+    )
+
+    # 4.3 is UKy Leaching - Tank Mixer
+    L_tank_mixer_accounts = ["4.3"]
+    m.fs.L_tank_mixers = UnitModelBlock()
+    m.fs.L_tank_mixers.power = Var(initialize=47.4, units=units.hp)
```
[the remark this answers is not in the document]
MarcusHolly: 4.74 hp just seemed a bit too low, but if you think 4.74 is a good value, I can fix that.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,839 @@ def display_results(m):
...
+    m = ConcreteModel()
+
+    # Add a flowsheet object to the model
+    m.fs = FlowsheetBlock(dynamic=True, time_units=units.s)
+    m.fs.costing = QGESSCosting()
+    CE_index_year = "UKy_2019"
+
+    # Leaching costs
+    # 4.2 is UKy Leaching - Polyethylene Tanks
+    L_pe_tanks_accounts = ["4.2"]
```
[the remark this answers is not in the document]
bpaul4: Hi @Agfritz, we implemented this approach in IDAES costing to better organize different cost accounts from different sources. For example, see Table 2 in this [IDAES costing documentation](https://idaes-pse.readthedocs.io/en/latest/reference_guides/model_libraries/power_generation/costing/power_plant_costing_netl.html) where there are a large number of accounts that cover subparts of plant sections. In the case of UKy, there are different types of accounts that have similar names and numbering helps differentiate them.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,839 @@ def display_results(m):
...
+            "n_equip": 2,
+            "scale_down_parallel_equip": False,
+            "CE_index_year": CE_index_year,
+        },
+    )
+
+    # 5.3 is UKy Rougher Solvent Extraction - Process Pump
+    RSX_pump_accounts = ["5.3"]
+    m.fs.RSX_pump = UnitModelBlock()
+    flow_5_3 = value(
```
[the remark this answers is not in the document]
MarcusHolly: The aqueous inlet is already at the pilot scale (rather than the commercial scale), so there's no reason to adjust this value lower.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,839 @@ def display_results(m):
...
+            "n_equip": 1,
+            "scale_down_parallel_equip": False,
+            "CE_index_year": CE_index_year,
+        },
+    )
+
+    # 9.3 is UKy Rare Earth Element Precipitation - Tank Mixer
+    reep_tank_mixer_accounts = ["9.3"]
+    m.fs.reep_tank_mixers = UnitModelBlock()
+    m.fs.reep_tank_mixers.power = Var(initialize=0.61, units=units.hp)
```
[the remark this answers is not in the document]
MarcusHolly: This is another one of the situations where I felt like it didn't make sense to reduce the power to 0.0061. However, I don't have a very strong intuitive sense of what the power of a tank mixer should be, so feel free to let me know what a reasonable value is.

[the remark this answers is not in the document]
andrewlee94: @Agfritz @bpaul4 We really need input from the costing team on good values for these. @MarcusHolly has experience with unit model development, but not in costing so we need guidance on these sorts of things.

src/prommis/uky/uky_flowsheet.py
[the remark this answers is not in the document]
Agfritz: Thank you for the explanation! That makes sense.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,839 @@ def display_results(m):
...
+            "n_equip": 1,
+            "scale_down_parallel_equip": False,
+            "CE_index_year": CE_index_year,
+        },
+    )
+
+    # 9.5 is UKy Rare Earth Element Precipitation - Filter Press
+    reep_filter_press_accounts = ["9.5"]
+    m.fs.reep_filter_press = UnitModelBlock()
+    m.fs.reep_filter_press.volume = Var(initialize=0.405, units=units.ft**3)
```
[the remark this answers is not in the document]
MarcusHolly: Same response here - I don't have a strong intuitive sense of what a filter press volume should be, but let me know if there's a value I should use instead.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -861,5 +873,839 @@ def display_results(m):
...
+        ),
+        units=units.kg / units.hr,
+        mutable=True,
+        doc="Product dysprosium mass flow",
+    )
+
+    m.fs.Gd_product = Param(
+        default=units.convert(
+            flowsheet.fs.roaster.flow_mol_comp_product[0, "Gd"]
+            * 362.5
```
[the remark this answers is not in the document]
MarcusHolly: Yes - I will add this as a dictionary

src/prommis/uky/uky_flowsheet.py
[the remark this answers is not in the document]
Agfritz: For a pilot scale I think 4.74 would be closer to the true value (likely something like these drum mixers: https://www.indco.com/shop/category/industrial-mixers/drum-mixers)
As we discussed, we could iterate on these in a future version.

src/prommis/uky/uky_flowsheet.py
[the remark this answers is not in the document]
Agfritz: This would correspond to the feedrate and kinetics of the precipitation tank. I have actually typically seen filter presses sized to the m**2. I think this assumption is ok for now (it seems reasonable for a bench or pilot scale).

src/prommis/uky/uky_flowsheet.py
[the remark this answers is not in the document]
Agfritz: Thanks @MarcusHolly and @andrewlee94 , I agree that 0.0061 is not reasonable for hp (I would assume it would be somewhere between 0.2-0.75) so I think this is reasonable for now.  This is another case where, in reality, the pilot team may just use whatever equipment they have available even if the sizing is not perfect. When we are scaling these up ideally the hp, tank size, and pump flow rate would all be linked and calculated in the unit model.

Agfritz (approved):
As we discussed, these interim initializations are good for now to demonstrate the costing. I will add an issue to integrate the costing with additional variables from the flowsheet in future iterations.

MarcusHolly:
@andrewlee94 This PR can be merged now, but there's two minor issues. (1) You requested changes earlier and (2) on the GitHub side of things, Alison is not a member of this repo yet so her approval doesn't count for anything.

andrewlee94 (approved):
Approving on the basis that remaining issues will be dealt with later.

MarcusHolly:
https://github.com/prommis/prommis/issues/50 outlines future improvements

Issue #40, opened 2024-02-27
Update Costing Tests Impacted By Pyomo Implicit Set Changes


bpaul4:
Recently, Pyomo upgraded to 6.7.1 which uses anonymous sets for `IndexedComponent` behavior rather than creating implicit sets. There are two costing tests that need to be updated to reflect this change.

bpaul4:
Closed per #41

PR #41, opened 2024-02-27
Fix Costing Tests Impacted By Pyomo Implicit Set Changes

bpaul4:
## Addresses Issue: 
#40 

- Add built-in exception when a costed block is re-costed
- Update tests

### Legal Acknowledgement

   employer.

### DIFF
--- a/src/prommis/uky/costing/ree_plant_capcost.py
+++ b/src/prommis/uky/costing/ree_plant_capcost.py
@@ -1369,7 +1369,13 @@ class QGESSCostingData(FlowsheetCostingBlockData):
                             ref_units[0][1]
                         ) / getattr(pyunits, ref_units[1])
                     except AttributeError:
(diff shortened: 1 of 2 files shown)


MarcusHolly (approved):

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -284,6 +284,12 @@ def build_process_costs(
         """
 
         # define costing library
+        if hasattr(self, "library") and self.library == "REE":  # costing already exists
+            raise Exception(
```
andrewlee94, line 288: 1. You should avoid using bare Exceptions (and pylint will complain about these). `RunTimeError` or maybe `AttributeError` would be better.
2. f-strings are generally faster than string formatting, so I would encourage you to use them instead of this.

andrewlee94 (changes requested):
Just a couple of minor requests.

andrewlee94 (approved):

PR #42, opened 2024-02-27
Dynamic SX model and flowsheet

ArkoprabhoDasgupta:
- Dynamic terms along with the graphical variation terms are added, to be reviewed
- Cost related equipment detail terms are mentioned in the model

### Legal Acknowledgement

   employer.


src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -156,3 +156,16 @@ def mass_transfer_term(b, t, s, k, l, m):
             self.mscontactor.stream_component_interactions,
             rule=mass_transfer_term,
         )
+
+        self.clean_sx_pe_tank_cap = Var(initialize = 1405, units=units.gal, doc="Volume of polyethylene tanks")
+        self.num_sx_pe_tank = Var(initialize = 5, doc="Number of polyethylene tanks")
```
andrewlee94, line 161: A minor comment, but I would not say "polyethylene" here - these tanks could be made of anything really.

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -156,3 +156,16 @@ def mass_transfer_term(b, t, s, k, l, m):
...
+
+        self.clean_sx_tank_mixer_pow = Var(initialize = 1, units=units.hp, doc="Power of tank mixers")
+        self.num_sx_tank_mixer = Var(initialize = 2, doc="Number of mixers")
+
+        self.clean_sx_process_pump_feed = Var(initialize = 281, units=units.gal/units.min, doc="Process pump feed rates")
+        self.num_sx_process_pump = Var(initialize = 3, doc="Number of process pumps")
+
+        self.clean_sx_mix_set_cap = Var(initialize = 2444, units=units.gal, doc="Volume of mixer settler tanks")
+        self.num_sx_mix_set = Var(initialize = 6, doc="Number of mixer settler tanks")
```
andrewlee94, line 170: Whilst these variables are nice, there is no point in adding them without having the associated correlations; otherwise these are just free variables with no defined value. Also, things like the pump generally should not be part of the unit model - there instead should be a separate pump in the flowsheet to capture this.

src/prommis/solvent_extraction/sx_model_flowsheet.py
```diff
@@ -7,20 +7,32 @@
...
 from prommis.solvent_extraction.ree_og_distribution import REESolExOgParameters
+from prommis.solvent_extraction.ree_aq_distribution import REESolExAqParameters
 from prommis.solvent_extraction.solvent_extraction import SolventExtraction
 
+import matplotlib.pyplot as plt
+
 m = ConcreteModel()
-m.fs = FlowsheetBlock(dynamic=False)
+
+time_duration = 60
```
andrewlee94, line 22: Could you create separate dynamic and steady-state tests please. We need to have both to ensure the model runs correctly in both cases, and dynamic tests tend to be computationally expensive so should be run sparingly (thus we want the steady-state version to run more frequently).

src/prommis/solvent_extraction/sx_model_flowsheet.py
```diff
@@ -7,20 +7,32 @@
...
 
+from idaes.core.util import DiagnosticsToolbox
+from pyomo.util.check_units import assert_units_consistent
+
 from prommis.leaching.leach_solution_properties import LeachSolutionParameters
 from prommis.solvent_extraction.ree_og_distribution import REESolExOgParameters
+from prommis.solvent_extraction.ree_aq_distribution import REESolExAqParameters
 from prommis.solvent_extraction.solvent_extraction import SolventExtraction
 
+import matplotlib.pyplot as plt
```
andrewlee94, line 18: Plotting functions generally should not be part of the unit model file. If necessary, these should be in separate utility modules, and even then there needs to be some demonstrable value (i.e. if it is just using matplotlib to plot a few easy to get trends, then that should be left to the user).

Note that testing plotting functions is also extremely difficult, which is part of why we avoid including it in the model files.

andrewlee94 (changes requested):

src/prommis/solvent_extraction/solvent_extraction.py
ArkoprabhoDasgupta: Okay, actually it was written this way in the phase 1 report actually that is why, then how shall I mention them only as tanks?

src/prommis/solvent_extraction/solvent_extraction.py
ArkoprabhoDasgupta: Okay, actually Marcus asked me to type these in the model, so that these variables need not be hardcoded every time and can be called from the model

src/prommis/solvent_extraction/sx_model_flowsheet.py
ArkoprabhoDasgupta: Fine I will create two separate tests in the model

src/prommis/solvent_extraction/sx_model_flowsheet.py
ArkoprabhoDasgupta: Okay I will remove the plots

src/prommis/solvent_extraction/solvent_extraction.py
andrewlee94: Just call them tanks. In the UKy process they happened to be made of polyethylene, but this model is general.

ArkoprabhoDasgupta: Okay

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -156,3 +156,23 @@ def mass_transfer_term(b, t, s, k, l, m):
             self.mscontactor.stream_component_interactions,
             rule=mass_transfer_term,
         )
+
+        self.clean_sx_pe_tank_cap = Var(
```
andrewlee94, line 160: I think all of this needs to be removed. Some of these already exist in the model elsewhere (number and volume of the tanks), some are not even part of the SX unit operation (e.g. pumps), and the rest are things that you do not have correlations for and should be added only as part of costing the flowsheet (e.g., mixer power).

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
```diff
@@ -7,20 +7,28 @@
 )
 from idaes.core.util.model_statistics import degrees_of_freedom as dof
 
+import matplotlib.pyplot as plt
```
andrewlee94, line 10: These imports should be removed - you are not using these anymore.

andrewlee94 (changes requested):
There are still a few things that need to be cleaned up here. Additionally, I think this PR should wait until the changes to support stripping are incorporated, as these two PRs will probably conflict with each other.

src/prommis/solvent_extraction/solvent_extraction.py
ArkoprabhoDasgupta: Okay I will remove them

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
ArkoprabhoDasgupta: I have removed them

src/prommis/solvent_extraction/solvent_extraction.py
ArkoprabhoDasgupta: I have removed them

ArkoprabhoDasgupta:
This needs to be re-reviewed.

ArkoprabhoDasgupta:
> There are still a few things that need to be cleaned up here. Additionally, I think this PR should wait until the changes to support stripping are incorporated, as these two PRs will probably conflict with each other.

So far the changes you have mentioned I have applied them.

ArkoprabhoDasgupta:
I would like to close this PR, as this has now become outdated. I have updated the dynamic model to be one that uses the leaching property package, and i am trying to debug it at present.

ArkoprabhoDasgupta:
To be closed for PR #66

PR #46, opened 2024-03-08
Support scrubbing and stripping in SX model

ArkoprabhoDasgupta:
## Addresses Issue: 
The old model was restricted only for the loading operation, where mass transfer was happening only from aqueous to organic phase.

## Summary/Motivation:
This new modification allows mass transfer from either of the two phases to the other. Scrubbing and stripping operations that involve REE transfer from organic to aqueous phase can be done here.

- A configuration called aqueous_to_organic was added with Bool domain. The default value is True, which allows loading, and the value False allows scrubbing and stripping.
- A test flowsheet added for stripping.

### Legal Acknowledgement

   employer.


src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
```diff
@@ -7,20 +7,28 @@
...
 m.fs.leach_soln = LeachSolutionParameters()
 
+number_of_stages=3
+
 m.fs.solex = SolventExtraction(
-    number_of_finite_elements=3,
+    number_of_finite_elements=number_of_stages,
     dynamic=False,
     aqueous_stream={
-        "property_package": m.fs.leach_soln,
```
MarcusHolly, line 23: Does this flowsheet still work if you use the leaching prop pack? I believe we opted to use it before since it was identical/near-identical to the SX aqueous prop pack and wanted to use the same prop pack that leaching was using for its aqueous solution

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -131,28 +125,103 @@ def build(self):
...
         self.mass_transfer_constraint = Constraint(
             self.flowsheet().time,
             self.mscontactor.elements,
             self.mscontactor.stream_component_interactions,
             rule=mass_transfer_term,
         )
+
+        self.clean_sx_pe_tank_cap = Var(
+            initialize=1405, units=units.gal, doc="Volume of polyethylene tanks"
+        )
```
MarcusHolly, line 211: As context, these are the SX parameters used for costing. I'm not sure if these should be Vars or mutable Params

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
ArkoprabhoDasgupta: U have to deactivate the reaction terms for the dynamic scenario, like set the terms like has_reactions to False.. I think it should work for steady state

MarcusHolly:
To resolve the black failure, you need to run `black [filepath]` for the 3 mentioned files, and you need to remove any unused imports in the files to resolve the pylint faliures.

MarcusHolly:
The UKy flowsheet tests are failing since we now need to specify the partition coefficients in the flowsheet file. The current implementation of the SX circuits in that flowsheet aren't accurate, but you should at least add this part so that the tests pass, and we can update how the flowsheet handles SX in a separate PR.

lbianchi-lbl:
- These changes are needed to enable the scrubbing & stripping functionality in the UKy flowsheet
- This is high priority but not for the March release

@MarcusHolly feel free to link any related issues, if any

src/prommis/solvent_extraction/ree_aq_distribution.py
```diff
@@ -158,5 +158,17 @@ def get_material_flow_terms(self, p, j):
                 to_units=units.mol / units.hour,
             )
 
+    def get_material_density_terms(self, p, j):
+        if j == "H2SO4":
```
andrewlee94, line 162: Why are you using H2SO4 here (and I note that earlier you refer to this as the solvent). This presumes that the concentration of H2SO4 is effectively constant (i.e. no dissociation) which is not correct.

The approach you are trying to use here works well-enough for the concentration of H2O in dilute aqueous solutions (where the concentration of H2O is essentially constant), but I do not think it is appropriate here (H2SO4 is not constant or the dominant species).

src/prommis/solvent_extraction/ree_og_distribution.py
```diff
@@ -144,5 +144,17 @@ def get_material_flow_terms(self, p, j):
                 to_units=units.mol / units.hour,
             )
 
+    def get_material_density_terms(self, p, j):
+        if j == "DEHPA":
+            return units.convert(
```
andrewlee94, line 149: The approach is probably suitable here, where the organic solvent dominates the system.

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -106,21 +115,8 @@ def build(self):
...
-                ("aqueous", "organic", "Y"): 99.9 / 100,
-                ("aqueous", "organic", "La"): 75.2 / 100,
-                ("aqueous", "organic", "Ce"): 95.7 / 100,
-                ("aqueous", "organic", "Pr"): 96.5 / 100,
-                ("aqueous", "organic", "Nd"): 99.2 / 100,
-                ("aqueous", "organic", "Sm"): 99.9 / 100,
-                ("aqueous", "organic", "Gd"): 98.6 / 100,
-                ("aqueous", "organic", "Dy"): 99.9 / 100,
-            },
             mutable=True,
```
andrewlee94, line 120: I think you should assign a default value here so that these are not left empty. I would suggest either a value of 1 or a small positive value.

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -131,24 +127,78 @@ def build(self):
...
-            phase_list = aqueous.phase_list
-            if s == b.mscontactor.elements.first():
-                return (
-                    b.mscontactor.material_transfer_term[t, s, (k, l, m)]
-                    == -b.mscontactor.aqueous_inlet_state[t].get_material_flow_terms(
-                        phase_list, m
-                    )
-                    * b.partition_coefficient[(k, l, m)]
-                )
+            if self.config.aqueous_to_organic == True:
```
andrewlee94, line 130: You can make this far more compact, and use most of the previous logic to do this. The only thing that changes is which stream you use for the distribution coefficient, so that is all that should be in the `if/else`.

I.e., revert this to how it was and start with:

```
if self.config.aqueous_to_organic:
    state = b.mscontactor.aqueous
    in_state = b.mscontactor.aqueous_inlet_state
else:
    state = b.mscontactor.organic
    in_state = b.mscontactor.organic_inlet_state

phase_list = state.phase_list
```

You can then use `state` and `in_state` in the constraint, and do not need to duplicate all the code.

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
andrewlee94: This should not be an issue, and if it is it suggests and issue elsewhere in your model. You should be able to have these equilibrium reactions in parallel with everything else.

If you have trouble solving the model using the leaching properties, try using the `DiagnosticsToolbox` to see what the issue is.

src/prommis/solvent_extraction/sx_stripping_flowsheet_steady.py
```diff
@@ -0,0 +1,107 @@
+from pyomo.environ import ConcreteModel, SolverFactory
```
andrewlee94, line 1: Rather than separate sub-flowsheets for each section of the SX system, could you create a single flowsheet that represents all parts of a single SX unit operation (loading, cleaning and stripping). This will be important both as a general example for users on how to model a full SX operation (e.g. it is easy for a user to take one of these flowsheets and assume it represents a full SX system when in fact it is only a part of the system), as well as for you to learn how to set up recycle streams.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,262 @@
...
+        m.fs.solex.mscontactor.organic[0, :].conc_mass_comp["Gd"].fix(1e-9)
+        m.fs.solex.mscontactor.organic[0, :].conc_mass_comp["Dy"].fix(1e-9)
+
+        m.fs.solex.mscontactor.organic_inlet_state[:].flow_vol.fix(62.01)
+
+        m.fs.solex.mscontactor.organic[0, :].flow_vol.fix(62.01)
+
+        return m
+
+    @pytest.mark.component
```
andrewlee94, line 150: AS this is a dynamic model, I think all of these should be `integration` tests rather than `component` tests. Integration is a tag we use to identify tests which will take a long time to run (such as dynamic simulations).

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,262 @@
...
+
+class TestSXmodel:
+    @pytest.fixture(scope="class")
+    def SolEx_frame(self):
+        m = ConcreteModel()
+
+        time_duration = 60
+
+        m.fs = FlowsheetBlock(
+            dynamic=True, time_set=[0, time_duration], time_units=units.hour
```
andrewlee94, line 35: Make sure you double check your results and ensure that all unit conversions are being done correctly (i.e. your time domain is in hours, but many unit and thermo variables are probably in seconds).

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,262 @@
...
+
+        m.fs.solex.mscontactor.organic[0, :].flow_vol.fix(62.01)
+
+        return m
+
+    @pytest.mark.component
+    def test_structural_issues(self, SolEx_frame):
+        model = SolEx_frame
+        dt = DiagnosticsToolbox(model)
+        dt.display_underconstrained_set()
```
andrewlee94, line 154: This test is not checking for anything - displaying output will not cause tests to fail. What you want here is `dt.assert_no_structural_warnings()`.

andrewlee94 (changes requested):

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,262 @@
...
+        ) == pytest.approx(144.92, rel=1e-1)
+        assert value(
+            m.fs.solex.mscontactor.organic[time_duration, 1].conc_mass_comp["Sm"]
+        ) == pytest.approx(26.18, rel=1e-1)
+        assert value(
+            m.fs.solex.mscontactor.organic[time_duration, 1].conc_mass_comp["Gd"]
+        ) == pytest.approx(12.315, rel=1e-1)
+        assert value(
+            m.fs.solex.mscontactor.organic[time_duration, 1].conc_mass_comp["Dy"]
+        ) == pytest.approx(7.17, rel=1e-1)
```
andrewlee94, line 262: GitHub appears to have lost my final comment: You should add a new test here to ensure there are no numerical warnings from the DiagnosticsToolbox (`assert_no_numerical_warnings()`).

src/prommis/solvent_extraction/sx_stripping_flowsheet_steady.py
MarcusHolly: I already have a version of this flowsheet in #47. It's just called `sx_model_flowsheet.py`. I created it while setting up the new UKy flowsheet, so I'm not sure if it'd be a good use of Arko's time to re-create this...

andrewlee94: I think it is an important step for Arko in order to learn how to do recycle streams himself. As you probably realize from our discussions, there are a lot of catches to be aware of and the best way to learn is to try doing it yourself.

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
ArkoprabhoDasgupta: For steady state, the leaching property package can be used, I was talking about the dynamic scenario

andrewlee94: I was referring to dynamics as well. I cannot see any reason why you should not be able to solve the dynamic model with the reactions present.

ArkoprabhoDasgupta: There was a DOF error if i remember correctly. I will recheck it

src/prommis/solvent_extraction/ree_aq_distribution.py
andrewlee94: No, I am not suggesting anything to do with dissociation. I am pointing out that you have made two assumptions here that are questionable, if not incorrect.

Firstly, the purpose of this method here is to determine how the accumulation term for each species will be calculated; i.e. this is effectively the concentration of each species. The reason we do it this way it to allow for cases where we do not have a concentration variable directly available. Thus, what you have effectively done here is say that the concentration of H2SO4 will be constant in the system and cannot change, which is almost definitely incorrect.

Now, you probably did this by copying other code where I had done this, but I suspect you missed a key point in those examples. They were all cases where we had dilute aqueous solutions, and this was only applied to the H2O in the system. For dilute aqueous solutions, we can assume that the concentration of the water dominates the system and is thus constant. However, this is only true when you consider a dilute system with a single bulk solvent.

Looking at your comments, you have stated that you are considering H2SO4 to be the solvent in this system, which is incorrect. The solvent is H2O still, and H2SO4 is a dissolved solute (which will likely vary in concentration). Also, unlike dilute aqueous systems, H2SO4 solutions also shows a distinct variation in density with concentration, thus the assumption of constant density is questionable (although probably OK for narrow ranges, which is what I did in the leaching properties).

Basically, you need to rethink how you are defining the system, and probably switch to what the leaching properties did.

ArkoprabhoDasgupta: Okay I get your point, so the solvent is H2O. I have to think about the H2SO4 concentration part

andrewlee94: Not really, you just need to realize that H2SO4 is a dissolved species and not H2O (line 49). Then, look at how many dissolved species you have versus how many total species, and think about why that is the case and what it means for writing the material balances.

The issue you have here is that you are copying something I did where I took some shortcuts without understanding what these shortcuts were or why I took them. You might want to step back and think about how you would write this yourself (in terms of equations, not code) first.

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -93,6 +93,15 @@ class SolventExtractionData(UnitModelBlockData):
         ConfigValue(domain=int, description="Number of finite elements to use"),
     )
 
+    CONFIG.declare(
+        "aqueous_to_organic",
+        ConfigValue(
+            default=True,
+            domain=Bool,
+            description="Direction of the transfer between two phases",
```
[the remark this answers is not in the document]
ArkoprabhoDasgupta: I actually meant the material transfer is happening from which phase to which phase. So can I write it as whether mass transfer is happening from aqueous phase (True) or organic phase (False) ?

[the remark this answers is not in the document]
andrewlee94: If this is meant to be setting the direction of transfer, then a `Bool` is not a suitable way to set this. You would instead want an `Enum`, but finding a concise way to name this would be hard.

Also, I think that is a bit of a misconception really - mass transfer in IDAES is bi-directional although a sign convention indicates which direction it is flowing. However, in this case the issue is less about which way it flows (which in many ways an internal thing) and more about how you are defining the mass transfer term mathematically.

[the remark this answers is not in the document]
ArkoprabhoDasgupta: Yes it is actually about the mass transfer term being defined on the basis of which phase

src/prommis/solvent_extraction/ree_aq_distribution.py
ArkoprabhoDasgupta: For now, I have switched to leaching property package for steady dynamics, and it is working fine. That package uses water as dominant component, so I think this is fine.

src/prommis/solvent_extraction/ree_og_distribution.py
ArkoprabhoDasgupta: Okay

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: This is only for the dynamic right? This is not for the steady state?

ArkoprabhoDasgupta: Okay

src/prommis/solvent_extraction/solvent_extraction.py
ArkoprabhoDasgupta: I have added 1 as a default value for the coefficients

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: I have updated

lbianchi-lbl:
@andrewlee94 @MarcusHolly any ideas on possible ways to address the current test failures?

ArkoprabhoDasgupta:
> 1. The failures related to dynamics should simply be removed from this PR (i.e. all the dynamic model files should be removed from this PR and handled in a subsequent PR)
> 2. The failures related to the UKy flowsheet will need to be updated the old-fashioned way by adjusting the solutions in the test file.

I have corrected the remaining errors

src/prommis/solvent_extraction/solvent_extraction.py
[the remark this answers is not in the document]
MarcusHolly: Andrew's comment on this has not been addressed, but I'm not sure that this alone should hold up the PR

MarcusHolly (approved):
LGTM with the exception of the minor comment that has not been addressed (using `Enum` instead of the bool config option)

PR #47, opened 2024-03-14
Fix SX Configuration in UKy Flowsheet

MarcusHolly:
## Summary/Motivation:
The initial implementation of the solvent extraction circuits was not accurate, so this PR updates the UKy flowsheet as well the example solvent extraction flowsheet to more closely match the REESim configuration.

### Legal Acknowledgement

   employer.


src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -131,28 +127,102 @@ def build(self):
...
+
+        self.clean_sx_process_pump_feed = Var(
+            initialize=281, units=units.gal / units.min, doc="Process pump feed rates"
+        )
+        self.num_sx_process_pump = Var(initialize=3, doc="Number of process pumps")
+
+        self.clean_sx_mix_set_cap = Var(
+            initialize=2444, units=units.gal, doc="Volume of mixer settler tanks"
+        )
+        self.num_sx_mix_set = Var(initialize=6, doc="Number of mixer settler tanks")
```
ArkoprabhoDasgupta, line 228: I think these terms were decided to not keep in the model, as they are not inherently a part of the model, and some of the terms are not required for SX. @andrewlee94 can you please confirm once?

andrewlee94: @ArkoprabhoDasgupta That should be fixed in you PR first, and then the changes merged in here. If we do it in two places at once, we will end up with a conflict.

ArkoprabhoDasgupta: Yes I noticed, I implemented those changes in a different branch from my PR, I will implement them

ArkoprabhoDasgupta: I have implemented the changes, please take a look at them

src/prommis/solvent_extraction/ree_aq_distribution.py
```diff
@@ -158,5 +158,17 @@ def get_material_flow_terms(self, p, j):
                 to_units=units.mol / units.hour,
             )
 
+    def get_material_density_terms(self, p, j):
+        if j == "H2SO4":
```
andrewlee94, line 162: This will get fixed by @ArkoprabhoDasgupta, but note that this makes some very big assumptions that are likely wrong. This will only affect dynamic simulations however.

andrewlee94 (commented):
A couple of comments, but I think we need to wait for @ArkoprabhoDasgupta's PR before we really take a look at this.

src/prommis/precipitate/tests/test_precipitator.py
```diff
@@ -247,8 +252,9 @@ def test_conservation(self, prec):
         )
 
         reversed_react = dict(map(reversed, prec.fs.properties_solid.react.items()))
+        bypass_elements = ["Ca", "H", "Cl", "SO4", "H2O", "HSO4"]
         for j in prec.fs.properties_aq.dissolved_elements:
-            if j == "Ca":
+            if j in bypass_elements:
```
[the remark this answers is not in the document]
MarcusHolly: Everything in `bypass_elements` does not have a corresponding oxalate term in `self.react` here: https://github.com/prommis/prommis/blob/5264136490bc1db391ef66ba13c0fcf6be391396/src/prommis/precipitate/precipitate_solids_properties.py#L76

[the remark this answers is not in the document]
andrewlee94: That is not quite what I was asking. My question was more "what happens to these materials"? Do they just end up in the liquid stream (i.e. do not precipitate and just pass throguh)?

If so, then you should still have a conservation check for these to ensure what went in came out in the liquid stream, but that is probably as simple as `assert value(flow_in) == pytest.approx(value(flow_out), rel=1e-5)`.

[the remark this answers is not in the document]
MarcusHolly: Yes, they just end up in the liquid stream. I've updated the test to ensure the difference between inlet and outlet concentrations for those elements are <= 1e-5

pyproject.toml
```diff
@@ -31,13 +31,13 @@ classifiers = [
 keywords = ["IDAES", "energy systems", "chemical engineering", "process modeling"]
 dynamic = ["version"]
 dependencies = [
-    "idaes-pse >= 2.3.0rc1",
+    "idaes-pse >= 2.4.0",
```
andrewlee94, line 34: Just to check that these are expected for this PR - these look like they really belong in a separate PR to update versions.

andrewlee94:
@MarcusHolly I suspect Github is confused, but they are showing up for me (check to see if you are displaying changes from all commits, or just since your last review).

MarcusHolly:
I'm seeing changes from all commits.

src/prommis/leaching/tests/test_leach_flowsheet.py
```diff
@@ -17,6 +22,10 @@ def model():
     m = build_model()
     set_inputs(m)
 
+    assert number_variables(m.fs.leach) == 197
```
andrewlee94, line 25: These should not be part of the fixture - it looks like this might have been carried over from before I updated the leaching code.

src/prommis/precipitate/tests/test_precipitator.py
```diff
@@ -247,9 +252,18 @@ def test_conservation(self, prec):
         )
 
         reversed_react = dict(map(reversed, prec.fs.properties_solid.react.items()))
+        pass_through_elements = ["Ca", "H", "Cl", "SO4", "H2O", "HSO4"]
         for j in prec.fs.properties_aq.dissolved_elements:
-            if j == "Ca":
-                pass
+            if j in pass_through_elements:
```
andrewlee94, line 257: Minor comment, but in this case it would be easier to write `value(a) == pytest.approx(value(b), rel=1e-5, abs=1e-5)`

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
```diff
@@ -14,120 +14,115 @@
...
+m.fs.solex.mscontactor.organic_inlet_state[0].conc_mass_comp["Ce"].fix(0.00031)
+m.fs.solex.mscontactor.organic_inlet_state[0].conc_mass_comp["Pr"].fix(3.711e-5)
+m.fs.solex.mscontactor.organic_inlet_state[0].conc_mass_comp["Nd"].fix(0.000165)
+m.fs.solex.mscontactor.organic_inlet_state[0].conc_mass_comp["Sm"].fix(1.701e-5)
+m.fs.solex.mscontactor.organic_inlet_state[0].conc_mass_comp["Gd"].fix(3.357e-5)
+m.fs.solex.mscontactor.organic_inlet_state[0].conc_mass_comp["Dy"].fix(8.008e-6)
+
+m.fs.solex.mscontactor.organic_inlet_state[0].flow_vol.fix(62.01)
+
+print(dof(m))
```
andrewlee94, line 106: We should avoid print statements like this in our code. Either add some text to tell the user what it was, or remove.

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
```diff
@@ -14,120 +14,115 @@
...
+
+dt = DiagnosticsToolbox(m)
```
andrewlee94, line 108: This looks like it should be in the tests rather than the flowsheet code.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -118,433 +131,633 @@ def test_build_flowsheet(self, model):
...
+        initialize_system(scaled_model)
 
-        assert degrees_of_freedom(model) == 0
+        assert degrees_of_freedom(scaled_model) == 0
 
     @pytest.mark.integration
-    def test_unit_consistency(self, model):
-        assert_units_consistent(model)
+    def test_unit_consistency(self, scaled_model):
+        assert_units_consistent(scaled_model)
```
andrewlee94, line 182: I would suggest replacing this with `assert_no_structural_warnings()` from the diagnostics toolbox.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -118,433 +131,633 @@ def test_build_flowsheet(self, model):
...
+        )
+        assert scaled_model.fs.precip_sx_mixer.outlet.flow_vol[
+            0
+        ].value == pytest.approx(5.76, 1e-4)
+        assert scaled_model.fs.precip_purge.inlet.flow_vol[0].value == pytest.approx(
+            0.0063, 1e-4
+        )
+
+    def test_main(self):
+        main()
```
andrewlee94, line 733: Does `main` have any meaningful assertions in it?

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -118,433 +131,633 @@ def test_build_flowsheet(self, model):
...
+        assert value(m.fs.costing.total_sales_revenue) == pytest.approx(
+            0.00018136, rel=1e-4
         )
 
     @pytest.mark.component
-    def test_report(self, model):
-        m = add_costing(model)
+    def test_report(self, scaled_model):
+        m = add_costing(scaled_model)
         display_costing(m)
```
andrewlee94, line 763: This does not look like it is actually asserting anything - i.e. all this tells us is that the code didn't raise an exception.

src/prommis/leaching/tests/test_leach_flowsheet.py
MarcusHolly: Yes, I believe this was carried over from before your changes. I'll remove it

src/prommis/solvent_extraction/sx_model_flowsheet_steady_state.py
MarcusHolly: Same comment as above @ArkoprabhoDasgupta

src/prommis/uky/tests/test_uky_flowsheet.py
MarcusHolly: I already have assert_no_strucutural_warnings in a previous test, so I suppose I can just delete this unit check since it's redundant.

src/prommis/uky/tests/test_uky_flowsheet.py
MarcusHolly: No - I had originally included it here for code coverage reasons, but I've since cleaned up the main function, so I don't believe this is necessary anymore (can't check the code coverage now due to the seemingly transient  testfailure)

src/prommis/uky/tests/test_uky_flowsheet.py
MarcusHolly: I added this just for code coverage.

andrewlee94: You need to assert something at least. For `display` type methods you should set up the `display` method to print to a user defined stream so that you can capture it here and assert that it is correct.

andrewlee94 (changes requested):
A few comments, the main ones being related to testing. A test has to assert something,and executing code just to meet code coverage requirements is not sufficient.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -118,389 +130,404 @@ def test_build_flowsheet(self, model):
...
 
-        assert degrees_of_freedom(model) == 0
+        # Solve scaled model
+        solve(scaled_model)
 
-    @pytest.mark.integration
-    def test_unit_consistency(self, model):
-        assert_units_consistent(model)
+        # Propagate results back to unscaled model
+        scaling.propagate_solution(scaled_model, model)
```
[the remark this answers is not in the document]
MarcusHolly: 1.) I've added `assert_optimal_termination` to this test

2.) I've based this test on what was done in `test_solve` of the `test_leach_flowsheet`, so I figured it'd be best to have the flowsheet tests be as consistent as possible. That being said, I could separate these out if we think that is better, although it might get a bit complicated with the scaling and propagation.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -216,28 +217,23 @@ def main():
...
+    display_results(m)
 
+    costing = add_costing(m)
     display_costing(costing)
 
-    return scaled_model, results
+    return m, results
 
 
 def build():
```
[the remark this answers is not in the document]
MarcusHolly: I will work on it for this PR

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -510,40 +537,111 @@ def test_solve_flowsheet(self, model):
...
+        assert model.fs.sl_sep2.recovered_liquid_outlet.flow_vol[
+            0
+        ].value == pytest.approx(6.3, 1e-4)
+        assert model.fs.precip_sep.inlet.flow_vol[0].value == pytest.approx(6.3, 1e-4)
+        assert model.fs.precip_sx_mixer.outlet.flow_vol[0].value == pytest.approx(
+            5.76, 1e-4
+        )
+        assert model.fs.precip_purge.inlet.flow_vol[0].value == pytest.approx(
+            0.63, 1e-4
+        )
```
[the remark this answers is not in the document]
MarcusHolly: This seems redundant since the precipitator unit model test already checks for conservation.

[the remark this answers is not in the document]
andrewlee94: You should check the full flowsheet still to be sure - there is the possibility of mistakes at the flowsheet level, and for small errors to accumulate.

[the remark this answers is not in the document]
MarcusHolly: Alright, I'll work on adding this then

src/prommis/uky/tests/test_uky_flowsheet.py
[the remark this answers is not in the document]
andrewlee94: They should be separated if possible, and if the leaching flowsheet is still like this after I reworked it then an issue should be opened to improve it.

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -193,6 +193,7 @@ def param_init(b, s, k, l, m):
             self.mscontactor.stream_component_interactions,
             initialize=param_init,
             mutable=True,
+            domain=Any,
```
andrewlee94, line 196: I do not think you should be using the `Any` domain - it is somewhat meaningless and I think it might trigger Pyomo warnings. If you need a domain, use `Reals` or `NonNegativeReals`.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -615,34 +595,104 @@ def test_solve_flowsheet(self, model):
         )
 
     @pytest.mark.unit
-    def test_costing(self, model):
-        m = add_costing(model)
+    def test_costing(self, system_frame):
+        model = add_costing(system_frame)
+        display_costing(model)
```
andrewlee94, line 600: Is this needed?

src/prommis/uky/uky_flowsheet.py
```diff
@@ -2020,16 +2013,17 @@ def function(stream):
...
-            initializer2.initialize(stream)
+            initializer_bt.initialize(stream)
 
     seq.run(m, function)
 
 
-def solve(m):
+def solve(m, solver=None):
     """
     Solve the system with IPOPT.
```
andrewlee94, line 2023: This doc string is not correct if you allow the user to provide a solver.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -2020,16 +2013,17 @@ def function(stream):
...
     """
-    solver = SolverFactory("ipopt")
+    if solver is None:
+        solver = SolverFactory("ipopt")
```
andrewlee94, line 2026: Should we use `get_solver` here as the default instead? This comes with some default options that might be useful, and might be important as we eventually switch to the new solver interface in Pyomo.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -2614,11 +2610,12 @@ def add_costing(flowsheet):
...
     # 4.2 is UKy Leaching - Polyethylene Tanks
+    tank_volume = value(
+        units.convert(flowsheet.fs.leach.volume[0, 1], to_units=units.gal)
+    )
     L_pe_tanks_accounts = ["4.2"]
     m.fs.L_pe_tanks = UnitModelBlock()
-    m.fs.L_pe_tanks.capacity = Var(
-        initialize=value(flowsheet.fs.leach.volume[0, 2]), units=units.gal
-    )
+    m.fs.L_pe_tanks.capacity = Var(initialize=tank_volume, units=units.gal)
```
andrewlee94, line 2618: Why are you using a constant value here, and not passing the leach tank volume directly? This means that the costing is static and will not be updated if you try to optimize the model for cost.

andrewlee94 (changes requested):
A few more comments.

ArkoprabhoDasgupta (approved):

src/prommis/solvent_extraction/solvent_extraction.py
MarcusHolly: If the domain is anything but `Any`, I see this error when trying to run any flowsheet with SX: 
image

Is the solution to this error specifying initial default values for each of the partition coefficients?

src/prommis/uky/tests/test_uky_flowsheet.py
MarcusHolly: No, I can delete it

src/prommis/solvent_extraction/solvent_extraction.py
andrewlee94: This should have `initialize=1` rather than calling a function that just assigns a value of 1 to any index. This should fix the issue.

@ArkoprabhoDasgupta The `initialize` argument needs to take either a single value, a `dict` with values for important keys, or an iterator - the function you provided does not actually return anything.

src/prommis/uky/tests/test_uky_flowsheet.py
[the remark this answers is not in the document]
MarcusHolly: In one of the recent commits, I've used the main() function as the fixture in the test file (which is done in some WaterTAP flowsheet tests), although this still does not address the "At the moment, the test fails if any step fails, but does not give you much information on where it failed". I've made a few attempts to separate the steps out, but it gets tricky with the scaling and propagation. I think we can update this (and rework the leaching flowsheet test later)

src/prommis/solvent_extraction/solvent_extraction.py
MarcusHolly: Thanks, having the following works:
`        self.partition_coefficient = Param(
            self.mscontactor.elements,
            initialize=1,
            mutable=True,
            doc="The fraction of component that goes from aqueous to organic phase",
        )`

src/prommis/uky/uky_flowsheet_ui.py
```diff
@@ -395,4 +397,17 @@ def get_diagram(build_options):
...
+    #
+    # scaled_model = set_scaling(m)
+    #
+    # initialize_system(scaled_model)
+    #
+    # results = solve(scaled_model)
+    #
+    # return results
+
     return solve(flowsheet)
```
MarcusHolly, line 413: @dangunter The UI test is failing (see screenshot), and I'm not quite sure how to resolve it. The issue seems to be that the solve function in the flowsheet has been updated in this PR to unfix and fix certain variables after the solve. The test will pass if I replace L413 with the commented code, but I imagine we'd need to make use of the `flowsheet` input argument, which the commented code does not do. Any thoughts on this?

src/prommis/uky/tests/test_uky_flowsheet.py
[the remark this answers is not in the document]
MarcusHolly: Added conservation checks for the REEs

MarcusHolly:
@andrewlee94 How should I go about ensuring that `conc_mass_comp` variables go to near-zero positive values (rather than near-zero negative values)? All of the tests pass locally, so I'm not able to replicate the failures...

andrewlee94:
@MarcusHolly How are you getting variables with values outside their bounds in the first place? This looks like it is happening in the initialization for the rougher mixer, so look at what your initial feed conditions are.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -44,655 +49,689 @@
...
+    m.results = res
+
+    return m
+
+
+@pytest.mark.component
+def test_structural_issues(system_frame):
+    model = system_frame
+    dt = DiagnosticsToolbox(model)
+    dt.report_structural_issues()
```
andrewlee94, line 70: Remove this from here - tests generally should not generate output.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -44,655 +49,689 @@
...
+    assert isinstance(model.fs.s20, Arc)
+    assert isinstance(model.fs.s21, Arc)
+    assert isinstance(model.fs.s22, Arc)
+    assert isinstance(model.fs.s23, Arc)
+    assert isinstance(model.fs.sep2_solid, Arc)
+    assert isinstance(model.fs.sep2_recovered_liquid, Arc)
+    assert isinstance(model.fs.s24, Arc)
+    assert isinstance(model.fs.s25, Arc)
+
+    assert degrees_of_freedom(model) == 0
```
andrewlee94, line 171: You don't need this - the diagnostics already does this.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -44,655 +49,689 @@
...
+
+        assert feed == pytest.approx(
+            filter_cake
+            + filter_cake_liquid
+            + load_purge
+            + scrub_purge
+            + precip_purge
+            + roaster_retained_liquid
+            + roaster_product,
+            rel=1e-4,
```
andrewlee94, line 735: This is a looser tolerance than I would normally like to see - does it pass with a tolerance of 1e-5?

src/prommis/uky/uky_flowsheet.py
```diff
@@ -2610,12 +2608,11 @@ def add_costing(flowsheet):
 
     # Leaching costs
     # 4.2 is UKy Leaching - Polyethylene Tanks
-    tank_volume = value(
-        units.convert(flowsheet.fs.leach.volume[0, 1], to_units=units.gal)
-    )
     L_pe_tanks_accounts = ["4.2"]
     m.fs.L_pe_tanks = UnitModelBlock()
-    m.fs.L_pe_tanks.capacity = Var(initialize=tank_volume, units=units.gal)
+    m.fs.L_pe_tanks.capacity = Var(
```
andrewlee94, line 2613: Why are you creating a new `Var` for this, and not just using the leach tank volume? This means the costing is disconnected from the flowsheet.

andrewlee94 (dismissed):
A few more requests/comments,

src/prommis/uky/tests/test_uky_flowsheet.py
MarcusHolly: It does not pass with a tolerance of 1e-5

andrewlee94: What is the difference, and do we have any idea where in the process the differences is arising?

MarcusHolly: I can look into where the difference is arising if we want to get this to pass with 1e-5 tolerance.
image

andrewlee94: Umm... you need to look into that. The relative error in those values is 1%, which is huge.

MarcusHolly: @andrewlee94 I've addressed this and changed the tolerance to 1e-7. I wasn't accounting for the dust stream in the roaster.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: @andrewlee94 @bpaul4 So to be clear, how should I handle using values from the flowsheet in the costing? For this particular example, it works if I write `m.fs.L_pe_tanks.capacity = flowsheet.fs.leach.volume[0, 1]`, but for other variables, I've encountered a few problems with this approach when it comes to the flow rates:

1.) In some cases, trying to set `m.fs.RSX_pump.feed_rate = units.convert(
            flowsheet.fs.solex_rougher_load.mscontactor.aqueous_inlet.flow_vol[0],
            to_units=units.gal / units.min,
        )`
results in the DOFs not being equal to zero since it forces me to unfix this line `m.fs.RSX_pump.feed_rate.fix()` as a result of the following error: `AttributeError: 'MonomialTermExpression' object has no attribute 'fix'`

2.) In other flow rate cases, I set the variable as done above and unfix the variable but no DOF errors occur (which doesn't really make sense to me)

3.) If I try to set the variables without doing the unit conversion (some variables do not require unit conversions), then I see the following error: 
`RuntimeError: Re-assigning the component 'flow_vol' from block 'fs.solex_rougher_load.mscontactor.aqueous_inlet_state[0.0]' to
block 'fs.RSX_pump' as 'feed_rate'.

This behavior is not supported by Pyomo; components must have a
single owning block (or model), and a component may not appear
multiple times in a block.  If you want to re-name or move this
component, use the block del_component() and add_component() methods.`

andrewlee94: I am nit sure what exactly is going on here, but what I can say is you are going about it the wrong way.

First `m.fs.L_pe_tanks.capacity = flowsheet.fs.leach.volume[0, 1]` will not work as expected; you cannot just attach a Pyomo component to a new object and expect it to work.

1. That error implies that whatever you are trying to fix is not a `Var` but and `Expression` or `Reference`. This will not work for obvious reasons. Additionally, you do not need an explicit convert when using `fix` or `set_value` - it will handle the conversion for you if you provide a value with units.
3. That is the result of what I said above - you cannot assign a Pyomo component to another Block.

However, this is mostly beside the point. There should be no need to create new Vars and Blocks here - you should just connect the costing to the existing unit model.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1629,7 +3053,7 @@ def add_costing(flowsheet):
     )
 
     feed_input = units.convert(
```
bpaul4, line 3055: On line 3365, the code includes:
```
m.fs.feed_input.fix()
m.fs.feed_grade.fix()
m.fs.recovery_rate_per_year.fix()
m.fs.solid_waste.fix()
m.fs.precipitate.fix()
m.fs.dust_and_volatiles.fix()
m.fs.power.fix()
```
These were added to the original costing tests because the variables were created from scratch and should be fixed during the full solve. However, these quantities should be calculated already by the UKy flowsheet, so it would be good to couple them with something like:

```
m.fs.feed_input = Var(initialize=reasonable_initial_guess, units=correct_units)
m.fs.feed_input_constraint = Constraint(expr=(m.fs.feed_input = units.convert(flowsheet_variable_or_expression, to_units=correct_units)))
```

for each of the variables. This will help ensure no DOF are changing, as well.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -16,535 +16,743 @@
 """
 
 from pyomo.network import Arc
-from pyomo.environ import value
-from pyomo.util.check_units import assert_units_consistent
```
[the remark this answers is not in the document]
MarcusHolly: Yes, it checks for unit consistency

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -16,535 +16,743 @@
...
+from pyomo.environ import (
+    assert_optimal_termination,
+    value,
+    units,
+    TransformationFactory,
+)
 
 from idaes.core import FlowsheetBlock
 from idaes.core.util.model_diagnostics import DiagnosticsToolbox
-from idaes.core.util.model_statistics import degrees_of_freedom
```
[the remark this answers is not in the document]
MarcusHolly: Yes, it checks for DOFs as well

src/prommis/uky/uky_flowsheet.py
```diff
@@ -212,32 +214,30 @@ def main():
     Run the flowsheet by calling the appropriate functions in series.
     """
     m = build()
+    set_partition_coefficients(m)
 
     set_operating_conditions(m)
 
     scaled_model = set_scaling(m)
+
     assert_units_consistent(scaled_model)
```
andrewlee94, line 223: Bare assertions are not good in user facing code. You should replace these with `if` statements that raise an appropriate error with a useful error message.

bpaul4 (commented):
@MarcusHolly the costing changes look good, thank you for implementing them. Just a couple other comments on the costing.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1020,103 +1607,431 @@ def initialize_system(m):
...
+        initializer_mix = MixerInitializer()
+        initializer_leach = LeachingTrainInitializer()
+        initializer_bt = BlockTriangularizationInitializer()
 
         propagate_state(m.fs.liq_feed)
         propagate_state(m.fs.sol_feed)
+        propagate_state(m.fs.org_feed)
+        propagate_state(m.fs.org_feed2)
 
         if stream == m.fs.leach_liquid_feed:
```
andrewlee94, line 1755: Why is this called "stream"? From the context, it appears that this is initializing unit operations thus the name is confusing me.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1020,103 +1607,431 @@ def initialize_system(m):
...
         initializer_feed = FeedInitializer()
         initializer_product = ProductInitializer()
-        initializer1 = MSContactorInitializer()
-        initializer2 = BlockTriangularizationInitializer()
+        initializer_sep = SeparatorInitializer()
+        initializer_mix = MixerInitializer()
+        initializer_leach = LeachingTrainInitializer()
+        initializer_bt = BlockTriangularizationInitializer()
 
         propagate_state(m.fs.liq_feed)
```
andrewlee94, line 1750: Why are these part of this function? For one, these are getting called at every step of the initialization, when for the most part these should not be changing. Secondly, these are affecting streams (Arcs), where as the SM iniitalizer works on unit models (and handles the propagation of state data internally).

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1020,103 +1607,431 @@ def initialize_system(m):
...
+                # Re-solve unit
                 solver = SolverFactory("ipopt")
                 solver.solve(m.fs.leach, tee=True)
                 # Unfix feed states
-                m.fs.leach_liquid_feed.flow_vol.unfix()
+                m.fs.leach.liquid_inlet.flow_vol.unfix()
                 m.fs.leach.liquid_inlet.conc_mass_comp.unfix()
                 m.fs.leach.solid_inlet.flow_mass.unfix()
                 m.fs.leach.solid_inlet.mass_frac_comp.unfix()
         elif stream == m.fs.leach_mixer:
```
andrewlee94, line 1803: This is a better way to do this. You could create lists of the units associated with each type of initializer and use `if stream in []`. That way, you need fewer `else` checks.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1020,103 +1607,431 @@ def initialize_system(m):
...
+            initializer_bt.initialize(stream)
 
     seq.run(m, function)
 
 
-def solve(m):
+def solve(m, solver=None):
     """
-    Solve the system with IPOPT.
+    Solve the scaled system.
```
andrewlee94, line 2023: It is not really accurate to say "scaled" model here - it will solve whatever model is passed in as `m`.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1020,103 +1607,431 @@ def initialize_system(m):
...
     """
-    solver = SolverFactory("ipopt")
+    if solver is None:
+        solver = get_solver()
     results = solver.solve(m, tee=True)
 
+    m.fs.rougher_org_make_up.outlet.flow_vol.unfix()
```
andrewlee94, line 2029: These seem out-of-place here; they are not related to solving the model, but rather changing the DoF for a later process.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1126,33 +2041,577 @@ def display_results(m):
...
+            units.convert(
+                m.fs.leach_solid_feed.flow_mass[0]
+                * m.fs.leach_solid_feed.mass_frac_comp[0, "Fe2O3"]
+                * metal_mass_frac["Fe2O3"],
+                to_units=units.kg / units.hr,
+            )
+        )
+    )
+
+    # total_ca_recovery = 100 * value(
```
andrewlee94, line 2197: Commented code.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1126,33 +2041,577 @@ def display_results(m):
...
+                metal_mass_frac["Al2O3"]
+                * m.fs.leach_solid_feed.outlet.mass_frac_comp[0, "Al2O3"]
+                * m.fs.leach_solid_feed.outlet.flow_mass[0],
+                to_units=units.kg / units.hr,
+            )
+        )
+        * 100
+    )
+
+    # ca_recovery = value(
```
andrewlee94, line 2384: Commented code.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1126,33 +2041,577 @@ def display_results(m):
...
+                * m.fs.leach_solid_feed.outlet.flow_mass[0],
+                to_units=units.kg / units.hr,
+            )
+        )
+        * 100
+    )
+
+    print(f"\nLeaching Aluminum recovery is {al_recovery} %")
+    print(f"Total aluminum recovery is {total_al_recovery} %")
+    # print(f"\nLeaching Calcium recovery is {ca_recovery} %")
```
andrewlee94, line 2573: Commented code.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -3117,26 +3121,26 @@ def add_costing(m):
...
-        m.fs.time, initialize=dust, units=units.ton / units.hr
-    )  # dust and volatiles
+        m.fs.time, initialize=9.5e-8, units=units.ton / units.hr
+    )
+    m.fs.dust_and_volatiles_constraint = Constraint(
+        expr=m.fs.dust_and_volatiles[0]
+        == units.convert(m.fs.roaster.flow_mass_dust[0], to_units=units.ton / units.hr)
+    )
+
     m.fs.power = Var(m.fs.time, initialize=14716, units=units.hp)
```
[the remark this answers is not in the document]
MarcusHolly: Tank mixers aren't explicitly modeled in this flowsheet, so I'm not sure how I would write this constraint since none of the unit models (to my knowledge) have variables for power or work.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1126,33 +2041,577 @@ def display_results(m):
...
-
-    # Add a flowsheet object to the model
-    m.fs = FlowsheetBlock(dynamic=True, time_units=units.s)
     m.fs.costing = QGESSCosting()
     CE_index_year = "UKy_2019"
 
     # Leaching costs
     # 4.2 is UKy Leaching - Polyethylene Tanks
     L_pe_tanks_accounts = ["4.2"]
     m.fs.L_pe_tanks = UnitModelBlock()
```
andrewlee94, line 2608: Why are you creating a new unit model for this, and not just appending to the existing unit model? I.e.:

```
m.fs.leach.costing = UnitModelCostingBlock(...)
```

Costing is intended to be a plug-in to the existing models, not something that stands separate. The only create new unit models for balance-of-plant equipment that isn't captured in the main flowsheet, and even then you can probably do a separate costing block attached to the flowsheet (without a corresponding UnitModleBlock).

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1528,31 +2960,23 @@ def add_costing(flowsheet):
...
-        costing_method_arguments={
-            "cost_accounts": R_conveyors_accounts,
-            "scaled_param": m.fs.R_conveyors.throughput,
-            "source": 1,
-            "n_equip": 1,
-            "scale_down_parallel_equip": False,
-            "CE_index_year": CE_index_year,
-        },
-    )
+    # 3.2 is UKy Roasting - Conveyors
```
andrewlee94, line 2965: Commented code - why is this being left out if other balance-of-plant equipment is included?

src/prommis/uky/uky_flowsheet.py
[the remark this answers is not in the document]
bpaul4: We add some power variables during costing, although they don't seem to add up to 14176 hp (L_tank_mixers 4.74 hp, RSX_tank_mixers 2.0 hp, CSX_tank_mixers 0.08 hp, reep_tank_mixers 0.61 hp). I could have missed some variables.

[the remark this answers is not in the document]
MarcusHolly: Ok, I've added the constraint setting it equal to the sum of these variables

src/prommis/uky/uky_flowsheet.py
MarcusHolly: So what was the consensus on this? Is it fine to keep `UnitModelBlock` for cases where balance-of-plant equipment aren't explicitly captured in the flowsheet? Or do we want to create generic unit models for these cases and directly add them to the flowsheet (tanks, process pumps, etc.). If the latter, that seems like it should be a part of a separate PR

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Changed to "unit"

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Moved the initializers outside of the function and created the aforementioned if loops.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Created a separate function for this and then re-solve the system

src/prommis/uky/uky_flowsheet.py
MarcusHolly: This particular case is being left out because `m.fs.roaster.flow_mass_product` introduces a warning (potential evaluation error) in the costing, so the tests don't pass. From what I can tell, it's because this mass flow needs a lower bound, but it's set up as an Expression rather than a Var. I've played around with ways to add bounds to Expressions, but I couldn't find any methods that worked. Converting it to a Var would allow us to set bounds, but that would likely require a decent number of changes to the roaster model, which is something @jinliangma would need to handle, I imagine.

andrewlee94: There is no way to add bounds to an Expression. What is the specific error message you see?

src/prommis/uky/uky_flowsheet.py
andrewlee94: Balance of plant equipment should stay as they are, but maybe add a comment with each to indicate this.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: There's a DiagnosticToolbox check to ensure there are no structural issues with the `add_costing` function, but there's a potential evaluation error with this one equipment. 
image

andrewlee94: There is a flag to get the test to skip potential evaluation errors. However, that also looks like something where there should be a bound added, so open a new issue and tag @jinliangma to fix that (and remove the flag to skip the evaluation errors when it is fixed).

andrewlee94: `assert_no_structural_issues(ignore_evaluation_errors=True)`

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -181,13 +182,20 @@ def test_solve(system_frame):
...
+    fix_organic_recycle(scaled_model)
+
     results = solve(scaled_model)
 
     scaling = TransformationFactory("core.scale_model")
     scaling.propagate_solution(scaled_model, model)
 
     assert_optimal_termination(results)
 
+    dt = DiagnosticsToolbox(model)
```
andrewlee94, line 196: Minor comment, but I would suggest making this a separate test.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1739,53 +1748,75 @@ def initialize_system(m):
...
+    bt_units = [
+        m.fs.leach_mixer,
+        m.fs.leach_sx_mixer,
+        m.fs.sl_sep1,
+        m.fs.roaster,
+    ]
+
+    def function(unit):
+        if unit in feed_units:
+            print(f"Initializing {unit}")
```
andrewlee94, line 1802: For consistency, I suggest using the logger instead of `print` for these so users can control how much output they want to see.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1973,8 +2001,8 @@ def function(stream):
                 m.fs.solex_cleaner_strip.mscontactor.aqueous_inlet_state[
                     0
                 ].conc_mass_comp.unfix()
-        elif stream == m.fs.precipitator:
-            print(f"Initializing {stream}")
+        elif unit == m.fs.precipitator:
+            print(f"Initializing {unit}")
             try:
                 initializer_bt.initialize(m.fs.precipitator)
             except:
```
andrewlee94, line 2008: Why do you need this `try/except` - why not just go to the `except` if you know the first pass will fail?

src/prommis/uky/uky_flowsheet.py
```diff
@@ -2005,35 +2033,34 @@ def function(stream):
...
-        elif stream == m.fs.precip_sep:
-            print(f"Initializing {stream}")
-            initializer_sep.initialize(m.fs.precip_sep)
-        elif stream == m.fs.precip_sx_mixer:
-            print(f"Initializing {stream}")
-            initializer_mix.initialize(m.fs.precip_sx_mixer)
         else:
-            print(f"Initializing {stream}")
-            initializer_bt.initialize(stream)
+            print(f"{unit} is not being initialized.")
```
andrewlee94, line 2037: What units are not being initialized, and why?

Also, this is still a lot of `if/else` statements. Is there a more efficient way you could do this? Also, seeing as the BT Initializer is more or less the default, would it be possible to have it be the blanket `else`?

src/prommis/uky/uky_flowsheet.py
```diff
@@ -2005,35 +2033,34 @@ def function(stream):
...
 
     seq.run(m, function)
 
 
 def solve(m, solver=None):
     """
-    Solve the scaled system.
+    Solve the model.
```
andrewlee94, line 2044: Need to include the arguments here and what in put is expected. It looks like `solver` is expected to be either a `solver` object or `None`, however this is not clear from the doc string. Also, this won't gracefully handle cases where `solver` is not a solver object.

Finally, you should let users control the `tee` option, and probably set it to `False` to limit output.

You might want to consider:

```
if hasattr(solver, "solve"):
    solver = solver
else:
    solver = get_solver(solver)

results = solver.solve(m, tee=tee)
```

src/prommis/uky/uky_flowsheet.py
```diff
@@ -1739,53 +1748,75 @@ def initialize_system(m):
     seq.set_guesses_for(m.fs.rougher_org_make_up.outlet, tear_guesses8)
     seq.set_guesses_for(m.fs.cleaner_org_make_up.outlet, tear_guesses9)
```
andrewlee94, line 1750: Above - this is a lot of tear streams. Do we actually need that many for this system? How many tear streams did the SM tool identify>

src/prommis/uky/uky_flowsheet.py
```diff
@@ -220,14 +218,25 @@ def main():
...
+    if degrees_of_freedom(scaled_model) != 0:
+        raise AssertionError(
+            "The degrees of freedom are not equal to 0."
+            "Check that the expected variables are fixed and unfixed."
+            "For more guidance, run assert_no_structural_warnings from the IDAES DiagnosticToolbox "
+        )
 
     initialize_system(scaled_model)
 
+    solve(scaled_model)
```
andrewlee94, line 230: I would rename this to `solve_system` to be consistent with `initialize_system`, and to avoid any confusion with the solver object.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -220,14 +218,25 @@ def main():
...
+
+    fix_organic_recycle(scaled_model)
```
andrewlee94, line 232: A comment is probably need here to clarify what this is doing, as you normally do not fix a recycle stream. I suspect this might be a case of a manual tear stream where you did not create an `Arc`, but you should explain that here.

src/prommis/uky/tests/test_uky_flowsheet.py
```diff
@@ -791,7 +797,7 @@ def test_costing(system_frame):
     )
     assert model.fs.costing.total_fixed_OM_cost.value == pytest.approx(7.2485, rel=1e-4)
     assert model.fs.costing.total_variable_OM_cost[0].value == pytest.approx(
-        10.7422, rel=1e-4
+        1.45469, rel=1e-4
```
andrewlee94, line 800: This is a rather big difference. What changes to cause this?

src/prommis/uky/uky_flowsheet.py
MarcusHolly: All of the units were being initialized. This was just a leftover check that I forgot to delete when making sure all units were initialized. I've made the BT Initializer the blank else statement now, but I'm not sure if there's a more efficient way to do this. 

The main issue is that the SolventExtraction models don't work with the BTInitializer or the MSContactorInitializer. Is the following error something wrong with the MSContactor model, the SolventExtraction model, or my flowsheet?
image

MarcusHolly: To my knowledge, the only DOFs I need to fix for the SX models in `set_operating_conditions` are the partition coefficients, so I'm not even sure what these 36 DOFs are.

src/prommis/uky/uky_flowsheet.py
MarcusHolly: Not all of these were necessary. I've narrowed it down to just the leaching unit and the SX units. Maybe the latter will be resolved with an `SXInitializer`?

src/prommis/uky/tests/test_uky_flowsheet.py
MarcusHolly: It's the result of changing `m.fs.power` from 14716 hp to 7hp. The former was a value carried over from the commercial scale costing parameters (I forget whether I had already scaled this down or not), and the latter addresses Brandon's previous [comment](https://github.com/prommis/prommis/pull/47#discussion_r1617233106).

agarciadiego (approved):
LGTM

A few comments that can be addressed in a different PR. I don't think this should stop this PR from getting merged.

The image of the flowsheet is really nice but could be easier to follow:

- Image should be consistent when showing mixers (we should show all)
- Add name of the mixers as all the other unit models
- Make stream naming a little more consistent or easier to follow (maybe add an extra term in name to show what part of the flowsheet we are on)
- Add product block
- Add comment on why we added eps term
- In the notebook we could add an explanation on the initialization loops

MarcusHolly:
> LGTM
> 
> A few comments that can be addressed in a different PR. I don't think this should stop this PR from getting merged.
> 
> The image of the flowsheet is really nice but could be easier to follow:
> 
> * Image should be consistent when showing mixers (we should show all)
> * Add name of the mixers as all the other unit models
> * Make stream naming a little more consistent or easier to follow (maybe add an extra term in name to show what part of the flowsheet we are on)
> * Add product block
> * Add comment on why we added eps term
> * In the notebook we could add an explanation on the initialization loops

Thanks, I will address these points in a separate PR. I've added these items to #13.

PR #23, opened 2023-12-21
Membrane examples using WaterTAP

mollydougher:
## Summary/Motivation:
Demonstrate critical mineral (lithium brine) separation using existing unit models (reverse osmosis, nanofiltration) in WaterTAP.

## Changes proposed in this PR:
- example using reverse osmosis to concentrate lithium
- example using nanofiltration to separate lithium and magnesium

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.md file
   at the top level of this directory.

2. I represent I am authorized to make the contributions and grant the license. If my employer has
   rights to intellectual property that includes these contributions, I represent that I have
   received permission to make contributions and grant the required license on behalf of that
   employer.

PR #49, opened 2024-03-25
Nanofiltration example flowsheet

mollydougher:
## Summary/Motivation:
Splitting PR #23 into two, only including the nanofiltration example for the March 2024 Release.

- Use the DSPM-DE nanofiltration model from WaterTAP to demonstrate a lithium brine separation of lithium and magnesium ions.

### Legal Acknowledgement

   employer.


mollydougher:
@andrewlee94 @agarciadiego @adowling2 This is ready to be reviewed. There are 4 pytest checks failing because of `test_sx_model.py`, which I am not sure how to best address.

andrewlee94:
@bpaul4 @ArkoprabhoDasgupta Could you take a look at the failing SX tests for us? I think the first issue here is that the tests will be very fragile due to the small magnitude values with a relative tolerance (even if the tolerance is very large). I would suggest this should use an absolute tolerance.

src/prommis/nanofiltration/nf_brine_plot.py
```diff
@@ -0,0 +1,498 @@
...
+        source=m.fs.unit.permeate, destination=m.fs.permeate.inlet
+    )
+    m.fs.nf_to_retentate = Arc(
+        source=m.fs.unit.retentate, destination=m.fs.retentate.inlet
+    )
+    TransformationFactory("network.expand_arcs").apply_to(m)
+    return m
+
+
+def fix_init_vars(m):
```
andrewlee94, line 197: I generally recommend no truncating function names to avoid any user confusion.

src/prommis/nanofiltration/nf_brine_plot.py
```diff
@@ -0,0 +1,498 @@
...
+def add_obj(m):
+    """
+    Adds objective to the pyomo model
+    """
+    # limit Li loss
+    m.fs.obj = Objective(
+        expr=m.fs.retentate.flow_mol_phase_comp[0, "Liq", "Li_+"], sense=minimize
+    )
+
+    # # maxmize the reduction in Mg:Li ratio
```
andrewlee94, line 246: Commented lines.

src/prommis/nanofiltration/nf_brine_plot.py
```diff
@@ -0,0 +1,498 @@
...
+    m.fs.recovery_con = Constraint(
+        expr=m.fs.unit.recovery_vol_phase[0.0, "Liq"] <= recovery_limit
+    )
+
+
+def optimize(m, solver):
+    """
+    Optimizes the flowsheet
+    """
+    print(f"Optimizing with {format(degrees_of_freedom(m))} DOFs")
```
andrewlee94, line 289: Again, I suggest using a logger and raising an exception rather than an assertion.

src/prommis/nanofiltration/tests/test_nf_brine.py
```diff
@@ -0,0 +1,48 @@
...
+from pyomo.environ import value
+
+import pytest
+
+pytest.importorskip("watertap", reason="WaterTAP dependency not available")
+# pylint: disable-next=wrong-import-position
+from prommis.nanofiltration.nf_brine import main
+
+
+@pytest.mark.requires_idaes_solver
```
andrewlee94, line 18: Just to check - does this mark exist? I have not seen it before, but it might be a PrOMMiS specific mark (although I do not see a pytest.ini file to declare them).

src/prommis/nanofiltration/tests/test_nf_brine.py
```diff
@@ -0,0 +1,48 @@
...
+            0.5077,
+        ],
+        "perm_ion_ratio": [
+            (m.fs.permeate.flow_mol_phase_comp[0, "Liq", "Mg_2+"].value / 0.024)
+            / (m.fs.permeate.flow_mol_phase_comp[0, "Liq", "Li_+"].value / 0.0069),
+            0.497,
+        ],
+    }
+    for model_result, testval in test_dict.values():
+        assert pytest.approx(testval, rel=1e-3) == value(model_result)
```
andrewlee94, line 48: Minor comment, but this is a fairly loose tolerance - solver tolerance should be 1e-6 so I normally use `rel=1e-5`.

andrewlee94 (commented):
Quite a few comments, but most are minor.

src/prommis/nanofiltration/tests/test_nf_brine.py
lbianchi-lbl: Good point, this is a marker from WaterTAP where we run pytest with non-IDAES Ipopt as part of the CI checks. Since that's not the case here, I suggest to remove it.

```suggestion
```

ArkoprabhoDasgupta:
> @bpaul4 @ArkoprabhoDasgupta Could you take a look at the failing SX tests for us? I think the first issue here is that the tests will be very fragile due to the small magnitude values with a relative tolerance (even if the tolerance is very large). I would suggest this should use an absolute tolerance.

Andrew, actually I went through this, this test file is the older one which I made, where there was only one single test file for the pytest, after that I made two separate files for the two scenarios of static dynamic as you mentioned, so I do not have this file any longer

ArkoprabhoDasgupta:
> > @bpaul4 @ArkoprabhoDasgupta Could you take a look at the failing SX tests for us? I think the first issue here is that the tests will be very fragile due to the small magnitude values with a relative tolerance (even if the tolerance is very large). I would suggest this should use an absolute tolerance.
> 
> Andrew, actually I went through this, this test file is the older one which I made, where there was only one single test file for the pytest, after that I made two separate files for the two scenarios of static dynamic as you mentioned, so I do not have this file any longer

The file which I have is the test_sx_model_steady.py which is already in my PR

andrewlee94:
@ArkoprabhoDasgupta I assume that new test file is till in one of your PRs however? If so, then those changes don't take effect until your PR gets merged so the question still remains as to why this is failing and whether it matters (probably not). Also, the comment I made about tolerances possibly applies in the new PR too.

ArkoprabhoDasgupta:
> @ArkoprabhoDasgupta I assume that new test file is till in one of your PRs however? If so, then those changes don't take effect until your PR gets merged so the question still remains as to why this is failing and whether it matters (probably not). Also, the comment I made about tolerances possibly applies in the new PR too.

So tell me one thing, this absolute tolerance part becomes decisive for the concentration values that are really small, so what values do you suggest I give for the bigger values and smaller values, because the absolute tolerance value will be really higher for the big values of concentrations

src/prommis/nanofiltration/README.md
```diff
@@ -0,0 +1,24 @@
+# membrane examples
```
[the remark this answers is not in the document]
mollydougher: I only see an `index.md` in the docs folder, should I create a new file `index.rst`? Also, does the text from this `README.md` file become the `index.rst`, or would I have a separate file for this, like `membrane_examples.rst`?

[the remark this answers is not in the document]
andrewlee94: @lbianchi-lbl Is probably the one to answer this - I am not sure if we have decided how the docs will be written yet (I just presumed it would be the same as IDAES).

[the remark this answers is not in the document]
lbianchi-lbl: That's a good question. I think that, for the time being (i.e. for the March release), this README would be just fine. Once the Sphinx harness is properly set up, then we can move the content appropriately.

lbianchi-lbl:
@mollydougher Now that #51 has taken care of the SX model failure, it looks like there's still one test failure (naively, it seems due to a tiny numeric discrepancy):

image

mollydougher:
> @mollydougher Now that #51 has taken care of the SX model failure, it looks like there's still one test failure (naively, it seems due to a tiny numeric discrepancy):
> 
> image

@lbianchi-lbl I have addressed this issue in commit 8858bbc. Do I need to update the headers in my `nf_brine.py`, `test_nf_brine.py`, and `nf_brine_plot.py` files prior to the merge?

lbianchi-lbl:
> > @mollydougher Now that #51 has taken care of the SX model failure, it looks like there's still one test failure (naively, it seems due to a tiny numeric discrepancy):
> > image
> 
> @lbianchi-lbl I have addressed this issue in commit [8858bbc](https://github.com/prommis/prommis/commit/8858bbcf13e9f0f63ea1f5d25f5724f9233c065f). Do I need to update the headers in my `nf_brine.py`, `test_nf_brine.py`, and `nf_brine_plot.py` files prior to the merge?

Great thanks, from my side you can leave them as-is. Usually we have an automated tool to write and check that each Python file has copyright information in the header, but I've just realized that we don't have that set up in this repository.

src/prommis/nanofiltration/nf_brine.py
```diff
@@ -0,0 +1,386 @@
...
+        m: pyomo model
+    """
+    solver = get_solver()
+    m = build()
+
+    initialize(m, solver)
+    _log.info("Initialization Okay")
+
+    if degrees_of_freedom(m) != 0:
+        raise Exception("Degrees of freedom were not equal to zero")
```
andrewlee94, line 62: It is generally considered bad practice to use a bare `Exception`, as it makes it harder to differentiate different types of failures (`except Exception:` catches basically everything). You should use more specific exception types wherever possible; here I suggest `ValueError`.

src/prommis/nanofiltration/nf_brine.py
```diff
@@ -0,0 +1,386 @@
...
+    Optimizes the flowsheet
+
+    Args:
+        m: pyomo model
+        solver: optimization solver
+    """
+    _log.info(f"Optimizing with {format(degrees_of_freedom(m))} DOFs")
+    simulation_results = solver.solve(m, tee=True)
+    if simulation_results.solver.termination_condition != "optimal":
+        raise Exception("The solver did not return optimal termination")
```
andrewlee94, line 289: As above.

src/prommis/nanofiltration/nf_brine.py
```diff
@@ -0,0 +1,386 @@
...
+        pressure_limit: upper bound on the outlet pump pressure
+    """
+    # bound the feed pressure to a reasonable value for nanofiltration
+    # choose an upper limit of 70 bar (https://doi.org/10.1021/acs.est.2c08584)
+    m.fs.pressure_constraint = Constraint(
+        expr=m.fs.pump.outlet.pressure[0] <= pressure_limit
+    )
+
+
+def optimize(m, solver):
```
andrewlee94, line 278: I think this name is slightly misleading - this is solving the model rather than strictly optimizing (i.e. it also works for deterministic problems).

src/prommis/nanofiltration/tests/test_nf_brine.py
```diff
@@ -0,0 +1,47 @@
...
+# pylint: disable-next=wrong-import-position
+from prommis.nanofiltration.nf_brine import main
+
+
+@pytest.mark.component
+def test_main():
+    """
+    Tests the execution of the main function in nf_brine.py
+    """
+    m = main()
```
andrewlee94, line 23: It might not be possible for this model (as it is rather complicated and comes from WaterTAP), but it would be nice to look at adding diagnostics checks in here too (`dt.assert_no_strucutral_warnings()` and `dt.assert_no_numerical_warnings()`).

This should probably wait until a later date however, as it will likely require some work.

andrewlee94 (commented):
A few minor comments from a quick read through.

lbianchi-lbl (approved):

PR #51, opened 2024-03-29
Ensure WaterTAP is imported if available and add workaround for SX model test failures

lbianchi-lbl:
### Legal Acknowledgement

   employer.

### DIFF
--- a/src/prommis/conftest.py
+++ b/src/prommis/conftest.py
@@ -3,7 +3,7 @@ import pytest
 
 # FIXME remove before merging
 try:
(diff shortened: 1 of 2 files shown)


lbianchi-lbl:
The only failing check is Codecov (2 lines), which is expected since the `except ModuleNotFoundError:` branch only applies to the non-developer installation where WaterTAP is missing, but where coverage is not collected.

MarcusHolly (approved):

bpaul4 (approved):

PR #71, opened 2024-04-30
Fix misnamed property in leaching package

andrewlee94:
## Summary/Motivation:
Whilst working on other code, I noticed that `dens_mass` was named `dens_mol` in the leaching properties. This PR fixes this.

- Fix misnamed property.
-

### Legal Acknowledgement

   employer.


agarciadiego (approved):
LGTM

ArkoprabhoDasgupta (approved):

bpaul4 (approved):

MarcusHolly (changes requested):
It seems like this naming error may have been copied over into other files as well, so it might be worth addressing here: https://github.com/prommis/prommis/blob/main/src/prommis/solvent_extraction/ree_aq_distribution.py

andrewlee94:
@MarcusHolly @ArkoprabhoDasgupta How do we want to go about fixing the misnamed property in the SX properties? Are there other PRs we should get merged first to avoid conflicts?

ArkoprabhoDasgupta:
> @MarcusHolly @ArkoprabhoDasgupta How do we want to go about fixing the misnamed property in the SX properties? Are there other PRs we should get merged first to avoid conflicts?

I think it would be fine, because the only thing that is changing is the definition of the density term of the constant term of the solvent that is water, I do not see any change in the way the state variables names, so I think it would be fine

MarcusHolly (approved):

andrewlee94:
@MarcusHolly @ArkoprabhoDasgupta I've renamed the properties in the SX model as well. I will note however that there were no tests for the density in the SX models, so we need to work on improving test coverage.

PR #74, opened 2024-05-01
Dynamic flowsheet of solvent extraction added

ArkoprabhoDasgupta:
> [!WARNING]
> This PR applies the same changes done in #85. We should wait until #85 is merged, and then solve any resulting conflicts here.

- Dynamic flowsheet added for solvent extraction
-Pytest for dynamic operation of flowsheet added

### Legal Acknowledgement

   employer.


src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,186 @@
...
+from prommis.solvent_extraction.solvent_extraction import SolventExtraction
+
+"""
+Method of building a dynamic solvent extraction model with a specified number of 
+stages and with two separate property packages for the two inlet streams.
+This is a loading operation, so no additional argument has to be specified.
+
+"""
+
+m = ConcreteModel()
```
andrewlee94, line 28: Could you put all this code into sub-methods, and add a `if __name__ == "__main__":` at the end. See the UKy or leaching flowsheet for an example. This makes it a lot easier to test and run the code.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,186 @@
...
+
+time_duration = 60
+
+m.fs = FlowsheetBlock(dynamic=True, time_set=[0, time_duration], time_units=units.hour)
```
andrewlee94, line 32: Make sure you are doing a unit consistency check on this flowsheet - I am not sure if the framework will do all the necessary unit conversion if units in the time domain do not match the units sued in the properties/unit models.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,186 @@
...
+
+"""
+Discretization of the time domain, and specification of the partition coefficients,
+volume, volume fractions, and the initial conditions of state variables for the components
+for all the stages. 
+
+"""
+
+m.discretizer = TransformationFactory("dae.finite_difference")
+m.discretizer.apply_to(m, nfe=5, wrt=m.fs.time, scheme="BACKWARD")
```
andrewlee94, line 63: 1. Is a backward finite difference method the correct one to chose for the time domain? I think you normally use a forward difference (implicit Euler) method for time domains.
2. Is 5 time points sufficient? That seems very few points to be using in an example.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,186 @@
...
+m.fs.solex.partition_coefficient[:, "aqueous", "organic", "La"] = 75.2 / 100
+m.fs.solex.partition_coefficient[:, "aqueous", "organic", "Ce"] = 95.7 / 100
+m.fs.solex.partition_coefficient[:, "aqueous", "organic", "Pr"] = 96.5 / 100
+m.fs.solex.partition_coefficient[:, "aqueous", "organic", "Nd"] = 99.2 / 100
+m.fs.solex.partition_coefficient[:, "aqueous", "organic", "Sm"] = 99.9 / 100
+m.fs.solex.partition_coefficient[:, "aqueous", "organic", "Gd"] = 98.6 / 100
+m.fs.solex.partition_coefficient[:, "aqueous", "organic", "Dy"] = 99.9 / 100
+
+"""
+Fixation of the inlet conditions and the initial state values for all the components.
```
andrewlee94, line 89: Typo: "Fixing inlet...", not "Fixation of the inlet ..."

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,186 @@
...
+
+"""
+Solution of the model and display of the final results.
+
+"""
+
+solver = SolverFactory("ipopt")
+solver.options["bound_push"] = 1e-8
+solver.options["mu_init"] = 1e-8
+solver.solve(m, tee=True)
```
andrewlee94, line 178: 1. Is there any dynamic behavior in this example? As far as I can tell there is no disturbance, thus this doesn't demonstrate much.
2. Do you need to be changing `bound_push` and `mu_init`? If you need to change these, it generally indicates that the model is poorly scaled.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,261 @@
+from pyomo.environ import (
```
andrewlee94, line 1: Need the copyright header.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,261 @@
...
+from prommis.leaching.leach_solution_properties import LeachSolutionParameters
+from prommis.solvent_extraction.ree_og_distribution import REESolExOgParameters
+from prommis.solvent_extraction.ree_aq_distribution import REESolExAqParameters
+from prommis.solvent_extraction.solvent_extraction import SolventExtraction
+
+solver = get_solver()
+
+
+class TestSXmodel:
+    @pytest.fixture(scope="class")
```
andrewlee94, line 28: Don't repeat the code here. You should use the methods I suggested you define in the model above here. Otherwise, there is no test for the actual flowsheet code.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,261 @@
...
+    @pytest.mark.component
+    def test_solve(self, SolEx_frame):
+        m = SolEx_frame
+        results = solver.solve(m)
+
+        # Check for optimal solution
+        assert check_optimal_termination(results)
+
+    @pytest.mark.integration
+    def test_solution(self, SolEx_frame):
```
andrewlee94, line 177: You should add a test using the diagnostics toolbox to `assert_no_numerical_warnings` here. I suspect this will likely show you a number of issues with this model that should be fixed.

andrewlee94 (changes requested):

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,261 @@
...
+        m = SolEx_frame
+        time_duration = 60
+        assert value(
+            m.fs.solex.mscontactor.aqueous[time_duration, 3].conc_mass_comp["HSO4"]
+        ) == pytest.approx(7.651e-5, rel=1e-1)
```
andrewlee94, line 182: This tolerance is way too loose - you are allowing a 10% variation in the value here. This should be rel=1e-4 at the most, and generally more like 1e-5.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: I will check it

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: 1. I will use Forward
2. I just gave 5 to check because it was going very long for more than 5 units.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: Okay

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: Okay I will check

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
andrewlee94: You need to check to be sure however.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
andrewlee94: Dynamic models always take a while to run. The number of time points needs to be matched to the time constant of the process.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: 2. I actually did not write this, this were given in the model when the prommis repo was being developed.
1. Can you elaborate the question? Are you asking whether there is any change with respect to time?

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: So Andrew, I saw the pytest code for the leaching flowsheet, are you asking me to write it in the same format?

andrewlee94: Yes. The point is to test all the code you have written, and if you duplicate the code all you are testing is the duplicated code. This then means the actual code is untested and can start to drift out of sync.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
andrewlee94: Yes - is there any change with time. If there isn't, you aren't really testing much,

ArkoprabhoDasgupta: There is change with time, in both the phases

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: Okay, I will do it

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: What do you mean by copyright header?

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: I have rechecked the equation, both sides of the material balance equation are mol/hour, regarding volume however I saw that the units are always converted to m3, even if I fix the units as L while fixing the volume. Not that it is affecting the overall equation in anyway

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: As discussed with you, I have given OCFE, and it is working, and the number of time points is more now.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
andrewlee94: You need to have a programmatic check (i.e. a test) to ensure this. The diagnostics toolbox can do this for you with the `assert_no_structural_warnings` method.

ArkoprabhoDasgupta: Okay I will see

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: I have given rel 1e-4 and an abs value of 1e-10

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: The assert_no_structural_warnings test is failing, there is a unit mismatch in the material accumulation term, where the DerivativeVar is being applied

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
andrewlee94: Did you try `rel=1e-5`? 1e-4 was my upper bound on the tolerance, and you should generally be able to get 1e-5 without issue.

Second, an `abs` tolerance of 1e-10 is way to tight to be meaningful (unless you have scaling) - solver tolerance is only 1e-8, so anything smaller than that will be lost to noise. Also, you generally only need to provide one of these (this is a test so you know the expected result and can determine a priori which tolerance will be the loosest).

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
andrewlee94: Good - now I know you actually did the test. To be sure however, you need to provide the traceback of the error message so we can be sure.

What I expect you are seeing is the result of an edge case in Pyomo, and unfortunately affects all DAE models with units. Basically, the issue arises in the numerical approximation of the derivative whee we get:

`der_var[t] == (v[t] - v[t-1])/([t] - [t-1])`

For this case, `der_var` has units of `mass/time` and `v` has units of `mass` (or `moles` as appropriate). Time should have units of `time`, which would make this consistent, but Pyomo `Sets` (and thus `ContinuousSets`) do not have units (semantically, as Set does not have units, although the elements inside it might). Unfortunately, from a coding sense, this is actually really hard to fix, so for now we have been living with the issue).

ArkoprabhoDasgupta: `pyomo.util.check_units:check_units.py:251 Units problem with expression fs.solex.mscontactor.organic_material_accumulation[60.0,3,liquid,Dy] - (0.499999999999981*fs.solex.mscontactor.organic_material_holdup[48.0,3,liquid,Dy] - 0.8285814518273972*fs.solex.mscontactor.organic_material_holdup[48.477718,3,liquid,Dy] + 0.6396718464360766*fs.solex.mscontactor.organic_material_holdup[50.376161,3,liquid,Dy] - 0.6860614273515556*fs.solex.mscontactor.organic_material_holdup[53.255698,3,liquid,Dy] + 0.9698922731138939*fs.solex.mscontactor.organic_material_holdup[56.345571,3,liquid,Dy] - 2.1365879070378115*fs.solex.mscontactor.organic_material_holdup[58.817579,3,liquid,Dy] + 1.5416666666667163*fs.solex.mscontactor.organic_material_holdup[60.0,3,liquid,Dy])`

ArkoprabhoDasgupta: unit assertion error

andrewlee94: I need the full output from the unit check to be certain. It should tell you what the unit inconsistency is too (i.e. units x are not compatible with units y). You need to use the Pyomo `assert_units_consistent` function for this. It would also help to know which constraint that is coming from.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: yes I tried, but there was assertion error after the 4th decimal place

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: `InconsistentUnitsError: Error in units found in expression: fs.solex.mscontactor.aqueous_material_accumulation[0.477718,1,liquid,H2O] - (-1.3155449435149056*fs.solex.mscontactor.aqueous_material_holdup[0.0,1,liquid,H2O] + 1.0466419563576284*fs.solex.mscontactor.aqueous_material_holdup[0.477718,1,liquid,H2O] + 0.3396521657396914*fs.solex.mscontactor.aqueous_material_holdup[2.376161,1,liquid,H2O] - 0.10143365070535666*fs.solex.mscontactor.aqueous_material_holdup[5.255698,1,liquid,H2O] + 0.04718598891201106*fs.solex.mscontactor.aqueous_material_holdup[8.345571,1,liquid,H2O] - 0.025592017688069048*fs.solex.mscontactor.aqueous_material_holdup[10.817579,1,liquid,H2O] + 0.009090500899000946*fs.solex.mscontactor.aqueous_material_holdup[12.0,1,liquid,H2O]): mole / hour not compatible with mole.`

andrewlee94: OK, so then it is the expected error in Pyomo. At least that means the original concern is not an issue.

ArkoprabhoDasgupta: So Andrew, what should I do next about this structural test fail? Because this will create a pytest fail everytime in git

andrewlee94: `assert_no_structural_warnings(ignore_unit_consistency=False)`

We have at least now confirmed that he only unit consistency issue if the expected one from the DAE tools ,so the best we can do now is to just ignore it.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
andrewlee94: There is a standard header that needs to go at the top of every file that states the copyright for the code. You can look at some of the other examples to see it.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: I think you meant `ignore_unit_consistency=True`

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: I have added

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,227 @@
...
+    m.fs.solex.mscontactor.organic[0, :].conc_mass_comp["Sm"].fix(1e-7)
+    m.fs.solex.mscontactor.organic[0, :].conc_mass_comp["Gd"].fix(1e-7)
+    m.fs.solex.mscontactor.organic[0, :].conc_mass_comp["Dy"].fix(1e-7)
+
+    m.fs.solex.mscontactor.organic_inlet_state[:].flow_vol.fix(62.01)
+
+    m.fs.solex.mscontactor.organic[0, :].flow_vol.fix(62.01)
+
+
+if __name__ == "__main__":
```
lbianchi-lbl, line 189: A possible way to address the missing test coverage for these lines:

1. Move the code in this `if` block into a function named e.g. `main()`
1. In the now empty `if` block, call the `main()` function (so that the effect of running this file will be the same)
1. In the appropriate test file, import the `main` function add a test function called e.g. `test_main_function()` that calls the `main()` function

The phrasing makes it sound more complicated that it actually is; basically, move the code into a function, and add a test that runs that function.

The same also applies for the steady-state version (and other files that follow the same `if __name__ == "__main__":` pattern)

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -86,8 +99,64 @@
...
+    """
+    This is a general purpose Initializer  for the Solvent Extraction unit model.
+
+    This routine calls the initializer for the internal MSContactor model.
+
+    """
+
+    CONFIG = ModularInitializerBase.CONFIG()
+
+    # CONFIG.declare(
```
andrewlee94, line 116: All this commented code should be removed - it is not applicable to modular initializers anyway.

ArkoprabhoDasgupta:
@andrewlee94 @agarciadiego @lbianchi-lbl @jinliangma i would like to request you to review this PR

ArkoprabhoDasgupta:
@lbianchi-lbl @agarciadiego @jinliangma please kindly give your review on this PR

src/prommis/solvent_extraction/solvent_extraction.py
```diff
@@ -4,6 +4,19 @@
...
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES).
+#
+# Copyright (c) 2018-2023 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
```
MarcusHolly, line 18: Why is the IDAES copyright header being added here?

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,295 @@
...
+    m.fs.solex.mscontactor.organic[0, :].flow_vol.fix(62.01)
+
+    return m
+
+
+@pytest.mark.unit
+def test_structural_issues(model):
+    dt = DiagnosticsToolbox(model)
+    dt.report_structural_issues()
+    dt.assert_no_structural_warnings(ignore_unit_consistency=True)
```
MarcusHolly, line 231: Why is the unit consistency being ignored? Is this just an ongoing issue with all dynamic flowsheets?

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,295 @@
...
+        "Nd": 0.9821,
+        "Pr": 0.21488,
+        "Sc": 1.76797,
+        "Sm": 0.10081,
+        "Y": 0.12888,
+    }
+
+    for k, v in model.fs.solex.organic_outlet.conc_mass_comp.items():
+        if k[0] == time_duration:
+            assert value(v) == pytest.approx(organic_outlet[k[1]], rel=1e-4)
```
MarcusHolly, line 291: Some changes have been made since the last time Andrew commented on this, so would this test pass now if you changed the relative tolerance to 1e-5?

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
```diff
@@ -0,0 +1,295 @@
...
+
+    for k, v in model.fs.solex.aqueous_outlet.conc_mass_comp.items():
+        if k[0] == time_duration:
+            assert value(v) == pytest.approx(aqueous_outlet[k[1]], rel=1e-4)
```
MarcusHolly, line 295: Same comment as above

src/prommis/solvent_extraction/solvent_extraction.py
ArkoprabhoDasgupta: I have removed it.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: There are derivative terms being included here, so during discretization, the unit of the denominator goes away. This issue is with any derivative term, as far as I know.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: The assertion test is failing at the 4th decimal place,  Obtained: 0.6020663510939577  Expected: 0.6021 ± 6.0e-06 , otherwise there is no big change in the values. Do you want me to update the values?

MarcusHolly: I don't think any changes are necessary here then

ArkoprabhoDasgupta: Okay

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: Same thing as above

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
dallan-keylogic: Could you please address this issue?

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
dallan-keylogic: Could you please fix this?

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,232 @@
...
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Fe"].fix(688.266)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Sc"].fix(0.032)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Y"].fix(0.124)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "La"].fix(0.986)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Ce"].fix(2.277)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Pr"].fix(0.303)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Nd"].fix(0.946)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Sm"].fix(0.097)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Gd"].fix(0.2584)
+m.fs.solex.aqueous_inlet.conc_mass_comp[:, "Dy"].fix(0.047)
```
dallan-keylogic, line 166: Where do these values comes from?

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
```diff
@@ -0,0 +1,232 @@
...
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Fe"] = 6.4 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Sc"] = 16.7 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Y"] = 99.9 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "La"] = 23.2 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Ce"] = 24.9 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Pr"] = 15.1 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Nd"] = 99.9 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Sm"] = 99.9 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Gd"] = 7.6 / 100
+        m.fs.solex.partition_coefficient[s, "aqueous", "organic", "Dy"] = 5 / 100
```
dallan-keylogic, line 142: Where do these values come from?

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
dallan-keylogic: I don't see the copyright header here.

dallan-keylogic (commented):
Now that Andrew is no longer on this project, I've been asked to step in and review this PR.

I see in several places you mentioned you fixed issues, then resolved the conversations, but the issues still aren't fixed in the PR. Could you have forgotten to commit them to Git or push your local commits to your Github branch?

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: I had applied this change some time back, there was an issue with code coverage in Git, so I had rewritten it. Andrew did not say anything about this after that.

src/prommis/solvent_extraction/tests/test_sx_model_dynamic.py
ArkoprabhoDasgupta: I have added it.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: It was taken quite some time back from the git prommis leaching flowsheet product stream values.

src/prommis/solvent_extraction/sx_model_flowsheet_dynamic.py
ArkoprabhoDasgupta: These values were taken from REESim file

ArkoprabhoDasgupta:
> Now that Andrew is no longer on this project, I've been asked to step in and review this PR.
> 
> I see in several places you mentioned you fixed issues, then resolved the conversations, but the issues still aren't fixed in the PR. Could you have forgotten to commit them to Git or push your local commits to your Github branch?

I might have missed out on some of the changes. I have replied to your comments and applied the changes as you said. Please check them once.

jinliangma:
I reviewed the unit model for the solvent extraction.  It seems that the model uses the MSContactor model in idaes base model library for handling the mass transfer between aqueous and organic phases. The only constraint in the solvent extraction unit model is to define the mass transfer term used by the MSContactor.  The only Param defined is the partition_coefficient, currently is mutable and specified by the user (flowsheet developer). To me the partition coefficient should be a function of local stage's composition rather than the a user specified parameter. In other words, it should be a Var. Since we don't have a good equilibrium model to calculate the partition coefficient, we still can fix that for each stage.
The definition of the partition coefficient is totally different from the partition coefficient typically defined in chemical engineering, which is [A]_org/[A]_aq, where [A] is the concentration of species A. In other words, the mass transfer rate is related to both the organic and aqueous phase leaving the stage. In current implementation, it seems that the mass transfer rate is related to the incoming aqueous or organic phase's composition. This approach is similar to the splitter fraction in the splitter model.  We probably need to revise the code such that the partition coefficient is consistent with typical chemical equilibrium definition.
The dynamic flowsheet models in both the unit model folder and Test folder for pytest may also need to be revised. Typical dynamic model testing requires imposing a step change or ramp change (or disturbance called in dynamic modeling literature).  It will be better to introduce a disturbance in the middle of the time period modeled.  Current test may impose a sudden step change at t=0 for the concentrations in the vessels of each stage.
Another possible issue is related to the material_accumulation and energy_accumulation terms in the MSContactor. Not sure how these terms are set for the initial condition.

ArkoprabhoDasgupta:
> I reviewed the unit model for the solvent extraction. It seems that the model uses the MSContactor model in idaes base model library for handling the mass transfer between aqueous and organic phases. The only constraint in the solvent extraction unit model is to define the mass transfer term used by the MSContactor. The only Param defined is the partition_coefficient, currently is mutable and specified by the user (flowsheet developer). To me the partition coefficient should be a function of local stage's composition rather than the a user specified parameter. In other words, it should be a Var. Since we don't have a good equilibrium model to calculate the partition coefficient, we still can fix that for each stage. The definition of the partition coefficient is totally different from the partition coefficient typically defined in chemical engineering, which is [A]_org/[A]_aq, where [A] is the concentration of species A. In other words, the mass transfer rate is related to both the organic and aqueous phase leaving the stage. In current implementation, it seems that the mass transfer rate is related to the incoming aqueous or organic phase's composition. This approach is similar to the splitter fraction in the splitter model. We probably need to revise the code such that the partition coefficient is consistent with typical chemical equilibrium definition. The dynamic flowsheet models in both the unit model folder and Test folder for pytest may also need to be revised. Typical dynamic model testing requires imposing a step change or ramp change (or disturbance called in dynamic modeling literature). It will be better to introduce a disturbance in the middle of the time period modeled. Current test may impose a sudden step change at t=0 for the concentrations in the vessels of each stage. Another possible issue is related to the material_accumulation and energy_accumulation terms in the MSContactor. Not sure how these terms are set for the initial condition.

So actually the solvent extraction dynamic model is actually an extension of the steady state model, that is why the parameters are defined in this manner, the parameters are the partition coefficients taken from the REESim file, and there the values are defined in a certain way that is different from the traditional definition of distribution coefficient. The point you have mentioned regarding partition coefficient as a function of the composition of the stage itself, that is something I have worked on, but it is not in this branch. I think that is something which can be done in a separate PR, because the steady state model has also to be changed in that way. Regarding the initialization, I have been using the steady state values to initialize the state variables through all the time points to the steady state value. The material accumulation terms values are set to 'None' during initialization.

andrewlee94:
There definitely needs to be a disturbance tested before this gets merged (along with diagnostics checks). The issues with dynamics only show up when you start changing things - things like stiffness, structural singularities and degeneracies, etc. only become apparent when things start changing, so a steady-state case is not enough to test a dynamic model.

The moment you do this, you will likely find the model fails to solve due to a number of assumptions inherent in the model (distribution coefficients, constant densities, etc.).

ArkoprabhoDasgupta:
> There definitely needs to be a disturbance tested before this gets merged (along with diagnostics checks). The issues with dynamics only show up when you start changing things - things like stiffness, structural singularities and degeneracies, etc. only become apparent when things start changing, so a steady-state case is not enough to test a dynamic model.
> 
> The moment you do this, you will likely find the model fails to solve due to a number of assumptions inherent in the model (distribution coefficients, constant densities, etc.).

So how do you suggest I modify the solvent extraction dynamic code? Do you want me to give like a step change disturbance in any of the variables feed values, and then check how subsequently the dynamics are going?

andrewlee94:
@ArkoprabhoDasgupta You need to change the test code, not the model code. What you want to do is start with the system at steady state and maintain that for at least two time steps (so three points steps including 0). Then introduce a moderate disturbance in one input variable and then have enough additional time steps to approach steady-state again (the closer you get the better idea you will have about stiffness). You then want to carefully examine the resulting response curve to make sure it is smooth, not instantaneous, non-oscillatory and that overall mass and energy are conserved (at a minimum).

To start, do one variable disturbance at a time, but you will want to try a number of disturbances, and then work up to doing multiple changes. I suggest starting with the total flow rate or the aqueous phase as the first disturbance, as I suspect you will quickly find issues.

dallan-keylogic:
To start, I would start at steady state, at a certain time have a step change in some inlet rare earth concentration, then view how the outlet rare earth concentration changes.

However, the issue Jinliang brought up is important---presently, material is partitioned between streams based on inlet flow rates instead of an equilibrium relationship between tank concentrations (represented in the outlet properties) in the aqueous and organic phases. That means the two phases will not be in equilibrium during the dynamics. We need to figure out if that's a major issue or not before proceeding.

ArkoprabhoDasgupta:
This has to be closed.

dallan-keylogic:
Closing this PR because it's superseded by #138 and any future dynamic flowsheet will be based on the model in #138.

PR #79, opened 2024-06-06
Add Diagnostics Tests For Costing

bpaul4:
## Summary/Motivation:

Applies DiagnosticsToolbox to costing tests and implements some minor modifications in the framework.

- Diagnostics checks in test module
- Some associated changes in the costing code itself

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.

### DIFF
--- a/src/prommis/uky/costing/ree_plant_capcost.py
+++ b/src/prommis/uky/costing/ree_plant_capcost.py
@@ -318,102 +318,102 @@ class QGESSCostingData(FlowsheetCostingBlockData):
             self.piping_materials_and_labor_percentage = Param(
                 mutable=True,
                 initialize=piping_materials_and_labor_percentage / 100,
-                doc="piping, materials and labor",
+                doc="Piping, materials and labor",
             )
 
             self.electrical_materials_and_labor_percentage = Param(
                 mutable=True,
                 initialize=electrical_materials_and_labor_percentage / 100,
-                doc="electrical, materials and labor",
+                doc="Electrical, materials and labor",
             )
 
             self.instrumentation_percentage = Param(
                 mutable=True,
                 initialize=instrumentation_percentage / 100,
-                doc="instrumentation",
+                doc="Instrumentation",
             )
 
             self.plant_services_percentage = Param(
                 mutable=True,
                 initialize=plants_services_percentage / 100,
-                doc="plant services",
+                doc="Plant services",
             )
 
             self.process_buildings_percentage = Param(
                 mutable=True,
                 initialize=process_buildings_percentage / 100,
-                doc="process buildings",
+                doc="Process buildings",
             )
 
             self.auxiliary_buildings_percentage = Param(
                 mutable=True,
                 initialize=auxiliary_buildings_percentage / 100,
-                doc="auxiliary buildings",
+                doc="Auxiliary buildings",
             )
 
             self.site_improvements_percentage = Param(
                 mutable=True,
                 initialize=site_improvements_percentage / 100,
-                doc="site improvements",
+                doc="Site improvements",
             )
 
             self.equipment_installation_percentage = Param(
                 mutable=True,
                 initialize=equipment_installation_percentage / 100,
-                doc="equipment installation",
+                doc="Equipment installation",
             )
 
             self.field_expenses_percentage = Param(
                 mutable=True,
                 initialize=field_expenses_percentage / 100,
-                doc="field expenses",
+                doc="Field expenses",
             )
 
             self.project_management_and_construction_percentage = Param(
                 mutable=True,
                 initialize=project_management_and_construction_percentage / 100,
-                doc="project management and construction",
+                doc="Project management and construction",
             )
 
             self.process_contingency_percentage = Param(
                 mutable=True,
                 initialize=process_contingency_percentage / 100,
-                doc="process contingency",
+                doc="Process contingency",
             )
 
             # ancillary cost variables
             self.ancillary_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="ancillary cost in $MM",
+                doc="Ancillary cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
             self.piping_materials_and_labor_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="piping, materials and labor ancillary cost in $MM",
+                doc="Piping, materials and labor ancillary cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
             self.electrical_materials_and_labor_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="electrical, materials and labor ancillary cost in $MM",
+                doc="Electrical, materials and labor ancillary cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
             self.instrumentation_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="ancillary cost in $MM",
+                doc="Ancillary cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
             self.plant_services_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="ancillary cost in $MM",
+                doc="Ancillary cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
@@ -421,28 +421,28 @@ class QGESSCostingData(FlowsheetCostingBlockData):
             self.buildings_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="buildings cost in $MM",
+                doc="Buildings cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
             self.process_buildings_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="process buildings cost in $MM",
+                doc="Process buildings cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
             self.auxiliary_buildings_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="auxiliary buildings cost in $MM",
+                doc="Auxiliary buildings cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
             self.site_improvements_costs = Var(
                 initialize=value(self.total_BEC),
                 bounds=(0, 1e4),
-                doc="site improvements buildings cost in $MM",
+                doc="Site improvements buildings cost in $MM",
                 units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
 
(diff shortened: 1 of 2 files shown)


bpaul4:
@andrewlee94 a known issue with installing WaterTAP is blocking the tests. In the meantime, I would appreciate your suggestions on the changes here and alternate/better ways to add Diagnostics to costing.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2497,6 +2497,7 @@ def calculate_REE_costing_bounds(
                 doc="feedstock capacity of site",
                 units=pyunits.tonnes,
             )
+            b.capacity.fix(value(pyunits.convert(capacity, to_units=pyunits.tonnes)))
```
andrewlee94, line 2500: You do not need the `value` or `convert` here - `b.capacity.fix(capacity)` will wokr as long as `capacity` is a Pyomo Var or Param or `pint` Quantity (which it appears to be).

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1109,6 +1123,19 @@ def test_REE_costing(self, model):
         # check that the model is set up properly and has 0 degrees of freedom
         assert degrees_of_freedom(model) == 0
 
+    @pytest.mark.component
+    def test_full_model_diagnostics(self, model):
+        dt = DiagnosticsToolbox(model)
+
+        try:
+            dt.assert_no_structural_warnings()
+        except AssertionError as e:
```
andrewlee94, line 1132: I would generally leave this out of the test - you have no idea exactly how the test might fail in the future so it is hard to predict what methods you might need to call for debugging. At most I would have `report_structural_issues` here as you don;t know what other methods might be useful.

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1150,7 +1194,7 @@ def test_results(self, model):
             65.333, rel=1e-4
         )
 
-    @pytest.mark.component
+    @pytest.mark.component  # TODO is this still needed?
```
andrewlee94, line 1197: No - the diagnostics test already includes this.

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1180,6 +1224,35 @@ def test_costing_bounding(self, model):
...
+            dt.report_numerical_issues()
+            dt.display_variables_at_or_outside_bounds()
+            dt.display_constraints_with_large_residuals()
+            dt.display_variables_with_extreme_jacobians()
+            dt.display_constraints_with_extreme_jacobians()
+            dt.display_near_parallel_constraints()
+            dt.display_near_parallel_variables()
+            raise (e)
+
         model.fs.costing.costing_lower_bound.pprint()
```
andrewlee94, line 1256: Print statements should not be included in tests - they slow things down.

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1180,6 +1224,35 @@ def test_costing_bounding(self, model):
             recalculate=True,
         )
 
+        dt = DiagnosticsToolbox(model)
+
+        try:
```
andrewlee94, line 1229: This is a lot of different `asserts` in a single test, and the whole test will fail at the first issues (thus hiding the other checks until the first one is fixes). Generally I would recommend having separate tests for each "type" of assertion or concept, so that you can get the full picture in one pass (rather than having to fix one issue only to find a new one).

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1206,9 +1279,14 @@ def test_costing_bounding(self, model):
         }
 
         for key in model.fs.costing.costing_lower_bound.keys():
-            assert model.fs.costing.costing_lower_bound[key].value == pytest.approx(
-                expected_costing_lower_bound[key], rel=1e-4
-            )
+            if model.fs.costing.costing_lower_bound[key].value <= 1e-4:
```
andrewlee94, line 1282: Why the `if/else`? `pytest.approx` can take both the `rel` and `abs` argument at the same time.

```
model.fs.costing.costing_lower_bound[key].value == pytest.approx(
    expected_costing_lower_bound[key],
    abs=1e-8,
    rel=1e-4,
)
```

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -905,6 +906,13 @@ def model(self):
         model = base_model()
         return model
 
+    @pytest.mark.component
+    def test_base_model_diagnostics(self, model):
+        dt = DiagnosticsToolbox(model)
+
+        dt.report_structural_issues()
```
MarcusHolly, line 913: Remove this function as tests shouldn't generally produce outputs

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1109,6 +1117,13 @@ def test_REE_costing(self, model):
         # check that the model is set up properly and has 0 degrees of freedom
         assert degrees_of_freedom(model) == 0
 
+    @pytest.mark.component
+    def test_full_model_diagnostics(self, model):
+        dt = DiagnosticsToolbox(model)
+
+        dt.report_structural_issues()
```
MarcusHolly, line 1124: Delete this

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1123,6 +1138,13 @@ def test_solve(self, model):
         results = solver.solve(model, tee=True)
         assert check_optimal_termination(results)
 
+    @pytest.mark.component
+    def test_solved_model_diagnostics(self, model):
+        dt = DiagnosticsToolbox(model)
+
+        dt.report_numerical_issues()
```
MarcusHolly, line 1145: Delete this

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1180,8 +1197,29 @@ def test_costing_bounding(self, model):
             recalculate=True,
         )
 
-        model.fs.costing.costing_lower_bound.pprint()
-        model.fs.costing.costing_upper_bound.pprint()
+        dt = DiagnosticsToolbox(model)
+
+        dt.report_structural_issues()
```
MarcusHolly, line 1202: Delete this

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1180,8 +1197,29 @@ def test_costing_bounding(self, model):
...
+        solver = get_solver()
+        results = solver.solve(model, tee=True)
+        assert check_optimal_termination(results)
+
+    @pytest.mark.component
+    def test_costing_bounding_solve_diagnostics(self, model):
+
+        dt = DiagnosticsToolbox(model)
+
+        dt.report_numerical_issues()
```
MarcusHolly, line 1218: Delete this

MarcusHolly (approved):
LGTM

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -1946,7 +1946,7 @@ def taxes_and_insurance_cost_rule(c):
         @b.Constraint()
         def sum_watertap_fixed_cost(c):
             if not hasattr(c, "watertap_fixed_costs_list"):
-                return c.watertap_fixed_costs == 0
+                return c.watertap_fixed_costs == 1e-12
```
andrewlee94, line 1949: Why are you changing this from 0 to such a small value?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2108,7 +2108,7 @@ def get_variable_OM_costs(
         )
 
         # assume the user is not using this
-        b.other_variable_costs.fix(0)
+        b.other_variable_costs.fix(1e-12)
```
andrewlee94, line 2111: Again, why fix this to such a small value?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2497,6 +2497,7 @@ def calculate_REE_costing_bounds(
                 doc="feedstock capacity of site",
                 units=pyunits.tonnes,
             )
+            b.capacity.fix(capacity)
             print("New variable 'capacity' created as attribute of {}".format(b.name))
```
andrewlee94, line 2501: You probably should be using a logger here.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2512,6 +2513,7 @@ def calculate_REE_costing_bounds(
                 doc="grade percentage of site",
                 units=pyunits.percent,
             )
+            b.grade.fix(grade)
             print("New variable 'grade' created as attribute of {}".format(b.name))
```
andrewlee94, line 2517: As above.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2539,7 +2541,7 @@ def calculate_REE_costing_bounds(
         if not hasattr(b, "costing_lower_bound"):
             b.costing_lower_bound = Var(
                 processes,
-                initialize=10,
+                initialize=1,
                 bounds=(0, 100),
                 doc="estimated lower bound on per unit production cost of site",
```
andrewlee94, line 2546: Typo: leading capital.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2559,7 +2561,7 @@ def calculate_REE_costing_bounds(
         if not hasattr(b, "costing_upper_bound"):
             b.costing_upper_bound = Var(
                 processes,
-                initialize=10,
+                initialize=1,
                 bounds=(0, 100),
                 doc="estimated upper bound on per unit production cost of site",
```
andrewlee94, line 2566: As above.

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1898,11 +1954,14 @@ def test_HDD_Recycling_costing_noOM_usedefaults():
         fixed_OM=False,
     )
 
+    dt = DiagnosticsToolbox(m)
```
andrewlee94, line 1957: This looks like it is one very large test - note that this can hide problems as the test will terminate at the first failure.

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -5265,13 +5412,16 @@ def test_REE_costing_recovery_transportcostParamnounits():
         transport_cost_per_ton_product=m.fs.transport_cost_per_ton_product,
     )
 
+    dt = DiagnosticsToolbox(m)
```
andrewlee94, line 5415: There appear to be a lot of very similar tests here - would it be possible to do this as a single parameterized test?

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: The diagnostics indicated an error from fixing the small value - the model can still solve with this specification.

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: See above.

src/prommis/uky/costing/ree_plant_capcost.py
andrewlee94: Was that an error, or just a caution?  This looks like a case where 0 should be a valid value, and thus you probably don't need to change this. The exception might be a possible evaluation error, but if that is the case it is more likely to be due to another unbounded variable.

bpaul4: You're correct, the check fails because those variables have lower bounds at 0. As these variables will always be fixed to 0 or a user input value, we could just remove the bounds and fix them to zero.

src/prommis/uky/costing/tests/test_ree_costing.py
bpaul4: I can parameterize parts of the file, particularly the parts that test different expressions, parameters, or variables for the same argument.

Is there a way to parameterize the entire file, or does too much differ between the tests?

andrewlee94: The parts where the tests are primarily the same (so the different expressions, parameters and variables, etc.). Some of the tests will be unique, but the point here is to minimize the amount of duplicated code that we need to maintain (which is a balancing act between having a small amount of highly re-used code that is hard to understand and maintain).

bpaul4: Unfortunately the way GitHub parsed the code difference is not very useful for reviewing. If you look at the file itself, I combined the options for (land_cost, additional_chemical_cost, additional_waste_cost) on line 2749, the options for recovery_rate units on line 3931, and the options for transport_cost on line 4137.

andrewlee94 (approved):

bpaul4:
@andrewlee94 codecov flagged 6 lines as "uncovered", which are all `_log(msg)` calls. Is this behavior expected?

andrewlee94:
@bpaul4 That indicates that you are not triggering those cases - generally speaking it is a good idea to try to trigger all 'if/else' paths, but often this ends up being not worth the effort. You should not aim for 100% coverage by lines - you should aim for 100% coverage by concept instead.

bpaul4:
@andrewlee94 all modified lines and "concepts" are covered, so I'd support merging this PR rather than adding line-by-line tests.

PR #84, opened 2024-06-26
Add Net Present Value Calculation to Costing Framework

bpaul4:
## Summary/Motivation:
Adds method to calculate the net present value (NPV), which is a better objective than cost of recovery to determine plant economic viability.

- Add method to calculate NPV from capital costs, operating costs, and revenue

- Operates as an optional cost framework option `build_process_costs([other arguments], calculate_NPV=True, [NPV arguments])` when both fixed and variable operating costs are calculated by referencing cost model variables and adding NPV calculations to the flowsheet costing block
- Also operations as a standalone method `QGESSCostingData.calculate_NPV(total_capital_cost, annual_operating_cost, annual_revenue, cost_year, [NPV_arguments])` that may be attached to any flowsheet block

- Supports a capital expenditure period, and independent cost escalation/inflation of capital costs, operating costs, and revenue via a calculated series present worth factor
- Supports royalties based on revenue or an optional user Expression for royalties
- Supports amortized loan interest based on a debt percentage of capital costs or an optional user Expression for debt
- Supports capital depreciation assuming a declining balance model
- Appropriate method docstrings and tests

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


bpaul4:
I will be adding more tests to cover the different use cases and switches, including the standalone case. In the meantime, please review the current code and tests. The pytests all pass locally.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2675,3 +2736,672 @@ def costing_upper_bound_eq(c, p):
 
         # method has finished building components
         b.components_already_built = True
+
+    def calculate_NPV(
+        # flowsheet block or costing block
+        b,
+        # required NPV arguments
+        discount_percentage,
```
andrewlee94, line 2744: This is a lot of arguments the user needs to provide to the method. This makes me think it would be better to define these in a `ConfigDict` on the class which the user can then set and play with more easily (and can be better documented), and this method can then access the values from the `ConfigDict`.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2675,3 +2736,672 @@ def costing_upper_bound_eq(c, p):
...
+            debt_expression: set the value or expression to calculate total debt
+            loan_interest_percentage: interest rate for loan repayment.
+            loan_repayment_period: length of repayment period in years.
+            capital_depreciation_declining_balance_percentage: factor to use for declining balance
+                depreciation. Set to 0 to indicate that capital does not depreciate over time.
+        """
+
+        # check if costing block was passed, and extract necessary components
+        # variables and constraints will be added to the costing block
+        if isinstance(b, FlowsheetCostingBlockData) and b.library == "REE":
```
andrewlee94, line 2928: To help with testing and maintenance, I would suggest breaking all of this input verification out into separate methods that you can unit test.

E.g. all this code for testing the type of the `Block` should be a separate method so that you can quickly test this in isolation by throwing different types of blocks at it (dummied up with and without the expected arguments) so that you can fully test all possible cases without needing to make sure the rest of the code works with the dummy cases.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2675,3 +2736,672 @@ def costing_upper_bound_eq(c, p):
...
+                REVENUE = costs["REVENUE"]
+        else:
+            # another block type is not allowed
+            raise TypeError(
+                f"Argument {b} is not a valid block type. The argument b "
+                "must be passed a FlowsheetCostingBlockData or FlowsheetBlock object."
+            )
+
+        # check capital expenditure arguments
+        if capital_expenditure_percentages is not None:
```
andrewlee94, line 3005: Similarity, with a `ConfigDict` for the arguments, all of this can be broken out into separate verification methods that can be tested in isolation.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2675,3 +2736,672 @@ def costing_upper_bound_eq(c, p):
...
+                    * CAPEX,
+                    to_units=cost_units,
+                )
+
+        else:
+
+            @b.Constraint()
+            def loan_debt_constraint(c):
+
+                return c.loan_debt == pyunits.convert(
```
andrewlee94, line 3329: Do you need the constraint in these cases? You are adding a lot of simple linear equality constraints to the model which is adding a lot of unnecessary burden on the solver (and writer).

I would suggest for these cases making `c.load_debt` a `Reference` based on the argument provided (noting you will need to move the unit conversion elsewhere).

andrewlee94: Similarly, you might want to break these sections out as separate sub-methods to make it easier to test them in isolation (so that you can set up a simple Block with the things you need for each branch of the `if/else` and make sure the variables/references and constraints are built correctly).

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -4434,3 +4465,98 @@ def test_REE_costing_recovery_Nonewithtransportcost():
...
+    assert hasattr(m.fs.costing, "operating_inflation_percentage")
+    assert hasattr(m.fs.costing, "revenue_inflation_percentage")
+    assert hasattr(m.fs.costing, "royalty_charge_percentage_of_revenue")
+    assert hasattr(m.fs.costing, "debt_percentage_of_CAPEX")
+    assert hasattr(m.fs.costing, "loan_interest_percentage")
+    assert hasattr(m.fs.costing, "loan_repayment_period")
+    assert hasattr(m.fs.costing, "capital_depreciation_declining_balance_percentage")
+
+    # check some NPV results
+    assert m.fs.costing.pv_capital_cost.value == pytest.approx(-14.33727, rel=1e-4)
```
andrewlee94, line 4554: Whilst this verifies the results of the calculations, this does very little to test that the variables and constraint were built as expected. I.e., this is a `component` or `integration` test, but you have no `unit` testing of the code.

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -4434,3 +4465,98 @@ def test_REE_costing_recovery_Nonewithtransportcost():
             ],
             transport_cost_per_ton_product=10,
         )
+
+
+@pytest.mark.component
+def test_REE_costing_NPV_defaults():
```
andrewlee94, line 4471: Building on all of this, you have a lot of `if/else` branches and possible exception paths in the code, but you are not testing any of these here. You need to have a lot more unit testing of these different paths, especially given how complex the code for this model is.

andrewlee94 (changes requested):

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: Thank you for the suggestion. I will look up the proper syntax for implementing this approach.

bpaul4: Would the new `CONFIG` variable belong to the `QGESSCostingData` class and include every argument, so I'd reformulate the other methods as well? I'm looking at code examples of implementations and deciding what I need to add to this file.

andrewlee94: No. You can overload the config from the parent class so only this class sees the changes.

That is in part why most model classes start with `CONFIG = CONFIG()` - note the `()`. By calling the existing `CONFIG` you are actually making a copy of it for the new class so that any changes you make do not affect the parent (but any changes in the parent will propagate down to the child).

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2669,3 +2858,737 @@ def costing_upper_bound_eq(c, p):
...
+            initialize=b.config.capital_depreciation_declining_balance_percentage,
+            units=pyunits.percent,
+        )
+
+        # define series present worth factor as an method so it can be called
+
+        def series_present_worth_factor(r, g, N):
+            """
+            Returns expression for series present worth factor.
+            """
```
MarcusHolly, line 3194:
```suggestion
            """
            Returns expression for series present worth factor where r is the discount rate
            expressed as a decimal, N is the project lifetime, and g is the escalation rate
            (e.g. inflation) expressed as a decimal.
            """
```
It might be worth adding this tidbit here as well for convenience

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2669,3 +2858,737 @@ def costing_upper_bound_eq(c, p):
...
+            def loan_debt_constraint(c):
+                # Debt  = - %debt_charge_of_CAPEX * CAPEX
+
+                return c.loan_debt == pyunits.convert(
+                    pyunits.convert(
+                        c.debt_percentage_of_CAPEX, to_units=pyunits.dimensionless
+                    )
+                    * c.CAPEX,
+                    to_units=c.cost_units,
+                )
```
MarcusHolly, line 3344: Should this equation have a negative sign in front?

bpaul4:
```suggestion
            def loan_debt_constraint(c):
                # Debt  = %debt_charge_of_CAPEX * CAPEX

                return c.loan_debt == pyunits.convert(
                    pyunits.convert(
                        c.debt_percentage_of_CAPEX, to_units=pyunits.dimensionless
                    )
                    * c.CAPEX,
                    to_units=c.cost_units,
                )
```

The equation is correct, but there shouldn't be a negative sign in the comment. The debt is taken as a positive value so that the interest owed (a negative cash flow) is `interest_owed = - (payment_per_year * number_of_years - debt)`.

MarcusHolly (approved):
LGTM

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2669,3 +2858,739 @@ def costing_upper_bound_eq(c, p):
...
+        and revenues to escalate over time via geometric gradient growth, e.g.
+        a constant proportional growth rate expressed as an escalation or
+        inflation percentage. Additionally, the formulation includes negative
+        cash flows from royalties, loan repayment and capital depreciation.
+        The general formulation is given by
+
+        NPV = PV_Revenue - PV_Operating_Cost - PV_Royalties - PV_Capital_Cost
+              - Loan_Interest_Owed - Capital_Depreciation
+
+        For escalating costs, the series present worth factor is modified to
```
Agfritz, line 2915: I would change this to say "for costs escalating at a constant rate for the project lifetime..." to help distinguish it from capital escalation

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2669,3 +2858,739 @@ def costing_upper_bound_eq(c, p):
...
+        is the percentage of the revenue charged as royalties expressed as a decimal.
+        The expressions above take the annuity growth during the entire analysis
+        period (NOp+NCap) and subtract the capital expenditure period (NCap) as there
+        is no operation or production during that time.
+
+        Finally, the full formulation supports estimation of interest
+        owed on loans and value losses due to capital depreciation. The loan
+        interest is calculated as an annual loan payment
+
+        Annual_Loan_Payment = Debt * [ iLoan_% * (1 + iLoan_%)**Nloan ] / [ (1 + iLoan_%)**Nloan - 1 ]
```
Agfritz, line 2965: I am a little bit confused about why the loan payment is included in the calculation. Typically, the interest is accounted for in the present worth factor. In the later calculations, it appears that you are already converting everything to present value. If this is the case, you do not need to include the loan interest in your summation for NPV (but please let me know if i am misinterpreting something in the formulation).

Agfritz (commented):
I have included mostly minor comments but did have one significant concern about the inclusion of loan interest in the final NPV summation. Please let me know if it would be helpful to discuss further.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -97,9 +98,151 @@ def custom_REE_plant_currency_units():
...
+            domain=bool,
+            description="True/false flag whether a capital expenditure period occurs.",
+        ),
+    )
+    CONFIG.declare(
+        "capital_expenditure_percentages",
+        ConfigValue(
+            default=None,
+            domain=list,
+            description="a list of values that sum to 100 representing how capital costs are spread over a capital expenditure period; for example, an input of [10, 60, 30] is parsed as a 3-year period where capital costs are spread as 10% in year 1, 60% in year 2, and 30% in year 3. The capital period precedes the operating period. Set to None to indicate no expenditure period.",
```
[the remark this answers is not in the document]
bpaul4: Yes, you are correct. No expenditure period means that 100% of the CAPEX occurs in first operating year t=0. An expenditure period of 1 year (e.g. `capital_expenditure_period=[100,]` would mean that "Year 1" is 100% of the CAPEX and the OPEX starts in the second year.

I'll update the description.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -1731,6 +1899,23 @@ def get_fixed_OM_costs(
                 f"{list(default_sale_prices.keys())}"
             )
 
+        # set default values
+        if labor_types is None:
```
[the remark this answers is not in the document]
bpaul4: The arguments are set using `None` as a guard value so that a new object is created every time the method is called and there is no chance of the object changing on each run.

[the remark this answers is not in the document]
Agfritz: Sounds good, thank you for clarifying Brandon!

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: I think you correct in your assessment. I understood that the interest should be included as a cost, but if it is already included in capital component then I can remove the loan repayment from the formulation. I'm happy to discuss it further here or offline.

bpaul4: The case study source provides a few rates - 3% inflation for operating costs, 3% inflation for revenue, 10% discount rate, 3.6% escalation rate for capital costs, and 6% interest rate on loans. Would it correct to use the 6% interest rate as the discount rate when projecting capital expenditure costs, with a 3.6% growth rate? And then 10% (in this scenario) would be the discount rate for operating costs, revenue and royalties, with a 3% growth rate.

bpaul4:
@Agfritz please re-review this at your convenience and let me know if you have further suggestions to improve the NPV formulation.

The last commit looks odd in the file comparison view but is actually a small change. I added a check

```
if (fixed_OM is False and calculate_NPV is True):  # NPV method invoked with fixed inputs
    self.calculate_NPV(fixed_OM, variable_OM)
else:  # continue on with building the plant costs
```

on line 457 to avoid creating unnecessary model variables and constraints when the user sets fixed values for the CAPEX, OPEX, and revenue to calculate the NPV. All of the code building the plant (overnight) costs, calling the O&M methods, etc is now indented under the `else`, and was otherwise not changed in any way.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2669,3 +2869,564 @@ def costing_upper_bound_eq(c, p):
...
+
+        b.pv_capital_cost = Var(
+            initialize=-b.CAPEX,
+            bounds=(-1e4, 0),
+            doc="present value of total lifetime capital costs in $MM; negative cash flow",
+            units=b.cost_units,
+        )
+
+        b.pv_operating_cost = Var(
+            initialize=-b.CAPEX,
```
Agfritz, line 3059: Should this be initialized as -b.OPEX? (similar question for initialization of other variables, not sure if it is important here)

Agfritz (approved):
These changes addressed my concerns. I have added two minor comments to clarify.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2669,3 +2869,564 @@ def costing_upper_bound_eq(c, p):
...
+                            to_units=pyunits.dimensionless,
+                        ),
+                        len(c.config.capital_expenditure_percentages),
+                    )
+                ),
+                to_units=c.cost_units,
+            )
+
+        if b.config.royalty_expression is None:
+            # PV_Royalties = - %royalty_charge_of_revenue * REVENUE
```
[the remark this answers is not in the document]
bpaul4: Yes, that is correct, I will update the comment.

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: CAPEX is used to initialize the PV values as a rough 1-2 order of magnitude initial guess based on user inputs. I agree that we can update the initial guesses, as there may be cases where CAPEX, OPEX and REVENUE do not have similar values.

ksbeattie:
@andrewlee94, can you re-review this?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -97,9 +98,144 @@ def custom_REE_plant_currency_units():
...
+            "capital expenses occur in the first operating year t=0.",
+        ),
+    )
+    CONFIG.declare(
+        "capital_escalation_percentage",
+        ConfigValue(
+            default=3.6,
+            domain=float,
+            description="rate at which capital costs escalate during the "
+            "capital expenditure period. Set to 0 to indicate expenditure is "
```
andrewlee94, line 191: "Set to 0 to indicate expenditure is spread but there is no cost escalation" implies there is an option for when the expenditure is not spread and there is no cost escalation. I am not sure this makes sense, in which case I would ask if the desired meaning is retained if "expenditure is spread but" is removed from the statement.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -97,9 +98,144 @@ def custom_REE_plant_currency_units():
...
+            "spread but there is no cost escalation.",
+        ),
+    )
+    CONFIG.declare(
+        "capital_loan_interest_percentage",
+        ConfigValue(
+            default=6,
+            domain=float,
+            description="interest rate for capital" "equipment loan repayment.",
```
andrewlee94, line 200: Remove the  " " from the middle of the doc string (this was caused by `black` removing an unneeded line break).

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -97,9 +98,144 @@ def custom_REE_plant_currency_units():
...
+        ),
+    )
+    CONFIG.declare(
+        "operating_inflation_percentage",
+        ConfigValue(
+            default=3,
+            domain=float,
+            description="inflation rate for operating costs during the "
```
andrewlee94, line 208: What are the units/basis for this? Given the values, I suspect this is a percentage value which needs to be stated explicitly.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -97,9 +98,144 @@ def custom_REE_plant_currency_units():
...
+            "operating period. Set to 0 to indicate no inflation.",
+        ),
+    )
+    CONFIG.declare(
+        "revenue_inflation_percentage",
+        ConfigValue(
+            default=3,
```
andrewlee94, line 215: As above, and also below.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -133,17 +269,10 @@ def build_process_costs(
...
-        labor_types=[
-            "skilled",
-            "unskilled",
-            "supervisor",
-            "maintenance",
-            "technician",
-            "engineer",
-        ],
-        labor_rate=[27.90, 23.26, 30.29, 24.06, 23.43, 46.82],
+        labor_types=None,
```
andrewlee94, line 272: Do you actually need all of these argument; they all look like things I just saw in the ConfigDict. If so, this seems like duplication of effect and an opportunity for something to go wrong. How do these arguments get used and populated?

Also, you should be able to remove the `# pylint: disable-next=dangerous-default-value` line above this.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -274,6 +404,34 @@ def build_process_costs(
             transport_cost_per_ton_product: Expression, Var or Param to use for transport costs
                 per ton of product (note, this is not part of the TOC)
             CE_index_year: year for cost basis, e.g. "2021" to use 2021 dollars
+            watertap_block: list of unit model blocks corresponding to watertap models
+            calculate_NPV: True/false flag for calculating net present value (NPV).
+
+            Keyword arguments related to NPV, entered at class instantiation:
```
andrewlee94, line 410: All of this should already be included in the main class doc string from the config block (and as mentioned above I think all these arguments are just duplicating effort).

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -296,648 +454,673 @@ def build_process_costs(
...
-                initialize=instrumentation_percentage / 100,
-                doc="Instrumentation",
-            )
+            # define variables
+            if Lang_factor is None:
+                # initialize parameters from specified percentages
+                self.piping_materials_and_labor_percentage = Param(
+                    mutable=True,
+                    initialize=piping_materials_and_labor_percentage / 100,
+                    doc="Piping, materials and labor",
```
andrewlee94, line 479: These doc strings do not really describe what this value is. Also, I note you are converting from a percentage value as an input to a decimal value for the parameter. I think this is a bad idea, as it creates a separation between the user input and the actual model which will result in confusion (i.e. a user might go to update one of these manually and expect to provide the value in the same basis they used when declaring the class).

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -296,648 +454,673 @@ def build_process_costs(
...
-                initialize=value(self.total_BEC),
-                bounds=(0, 1e4),
-                doc="Buildings cost in $MM",
-                units=getattr(pyunits, "MUSD_" + CE_index_year),
-            )
+                self.instrumentation_costs = Var(
+                    initialize=value(self.total_BEC),
+                    bounds=(0, 1e4),
+                    doc="Ancillary cost in $MM",
+                    units=getattr(pyunits, "MUSD_" + CE_index_year),
```
andrewlee94, line 568: Rather than build the units string each time you use it, just build it once up front as a local object and use it wherever it is needed. Otherwise, you are duplicating operations and wasting CPU time.

Also, what happens if the user wants to use Euros or some other currency units?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -296,648 +454,673 @@ def build_process_costs(
...
-            self.process_buildings_costs = Var(
-                initialize=value(self.total_BEC),
-                bounds=(0, 1e4),
-                doc="Process buildings cost in $MM",
-                units=getattr(pyunits, "MUSD_" + CE_index_year),
-            )
+                self.plant_services_costs = Var(
+                    initialize=value(self.total_BEC),
+                    bounds=(0, 1e4),
+                    doc="Ancillary cost in $MM",
```
andrewlee94, line 574: Minor comment, but I dislike $MM as this is a very US-centric definition (outside the US a single M indicates millions). Also, there is no need for the currency basis to be dollars - it could well be Euros or GBP, etc.

Thus, I would recommend a more general definition of the units in the doc string.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -296,648 +454,673 @@ def build_process_costs(
...
-            self.auxiliary_buildings_costs = Var(
-                initialize=value(self.total_BEC),
-                bounds=(0, 1e4),
-                doc="Auxiliary buildings cost in $MM",
-                units=getattr(pyunits, "MUSD_" + CE_index_year),
-            )
+                # buildings cost variables
+                self.buildings_costs = Var(
+                    initialize=value(self.total_BEC),
+                    bounds=(0, 1e4),
```
andrewlee94, line 581: How reasonable are these bounds, especially given that the basis and units of cost are not well known?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -296,648 +454,673 @@ def build_process_costs(
...
-            )
+                def contingency_cost_rule(self):
+                    return self.contingency_costs == (self.process_contingency_costs)
 
-            def contingency_cost_rule(self):
-                return self.contingency_costs == (self.process_contingency_costs)
+                self.contingency_cost_eq = Constraint(rule=contingency_cost_rule)
 
-            self.contingency_cost_eq = Constraint(rule=contingency_cost_rule)
+                def total_installation_cost_rule(self):
```
andrewlee94, line 816: A minor suggestion, but the `@m.Constraint` decorator generally results in more compact and readable code.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -296,648 +454,673 @@ def build_process_costs(
...
             else:
-                if pyunits.get_units(land_cost) == pyunits.dimensionless:
-                    self.land_cost = Expression(expr=land_cost * CE_index_units)
+                self.land_cost = Expression(expr=0 * CE_index_units)
+
+            # define feed input, if passed
+            if feed_input is not None:
+                if (
+                    pyunits.get_units(feed_input) == pyunits.dimensionless
+                ):  # assume it's short ton per hour
```
andrewlee94, line 875: Is this a good assumption? Generally when there is any ambiguity (or in this case an outright unit inconsistency) I generally recommend raising an Exception instead of making an assumption (which is quite likely wrong).

andrewlee94 (changes requested):
There is a lot to work through here, but here are some comments so far.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -97,9 +98,144 @@ def custom_REE_plant_currency_units():
...
+    CONFIG.declare(
+        "discount_percentage",
+        ConfigValue(
+            default=None,
+            domain=float,
+            description="rate of return used to discount future cash flows "
+            "back to their present value. The NETL QGESS recommends setting "
+            "the discount rate as the calculated after-tax weighted average "
+            "cost of ccapital (ATWACC). For the UKy case study, the ATWACC is"
+            "5.77%.",
```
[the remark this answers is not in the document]
bpaul4: For 1, Users should specify the discount percentage and project lifetime, as these two inputs have an outsized impact on the NPV calculation and are project-specific. I think it better to enforce that users set values for these two inputs and include the comment about the typical value rather than have defaults.

I'll update the doc strings to address 2 and 3. All percentage inputs are assumed to be percentage values (e.g. 5.77 means 5.77%).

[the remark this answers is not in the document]
andrewlee94: Regarding 1 then, I would remove the comment about the value for the UKy case study, as it is too-specific for a general tool.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -97,9 +98,144 @@ def custom_REE_plant_currency_units():
...
+            domain=bool,
+            description="True/false flag whether a capital expenditure period "
+            "occurs.",
+        ),
+    )
+    CONFIG.declare(
+        "capital_expenditure_percentages",
+        ConfigValue(
+            default=None,
+            domain=list,
```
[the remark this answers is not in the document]
bpaul4: I see that `ListOf` supports arguments `itemtype`, `domain`, and `string_lexer`. Is there a way to use `domain` or `string_lexer` to validate something like `ListOf(float, domain=values_sum_to_100)`?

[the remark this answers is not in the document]
andrewlee94: I doubt it, but you would need to read the docs.

src/prommis/uky/costing/ree_plant_capcost.py
andrewlee94: My point is that these already exist in the ConfigDict do they not? If that is the case, why create a new local variable for something that is already stored elsewhere.

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: For millions of USD, the scale of plants should not exceed 10,000 million USD for any of the cost variables so that should be a sufficiently large upper bound. The supported USD cost years have index values between ~350 to 850, so the upper bound works for any cost year.

We would probably want to change or remove the upper bound if we're allowing cost units other than "millions of USD".

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: Addressing this comment and the one above, we don't have a dictionary of cost index values for non-USD currency units. However, users may add their own custom units by registering them, for example:

```
pyunits.load_definitions_from_strings(
        [ "custom_currency_units= 500/700 * USD_CE500",]
    )
```

where `custom_currency_units` is effectively a CE index value of 700, but really this is currency conversion rate between USD_CE500 and the new units. If users know the conversion, say 1.05 euro to 1 dollar in 2022, they could do something like

```
pyunits.load_definitions_from_strings(
        [ "EUR_2022= 1/1.05 * USD_2022",]
    )
```

We would need to update the code, variables bounds, doc strings, etc. to account for other units besides USD.

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: See answer below.

src/prommis/uky/costing/ree_plant_capcost.py
andrewlee94: Actually, a better answer would be to never write units into the doc strings and just store them as units on the object as this gets around the entire issue and avoids duplicating information. We should then make use of the fact that pint supports the concept of `Munits = 1e6*units` to handle the fact that this is in millions of currency units (also as a note for the future, some currencies would want different multipliers due to smaller or larger units of currency).

src/prommis/uky/costing/ree_plant_capcost.py
andrewlee94: The CEI values have been steadily increasing over the years however, so eventually they will reach a point where the bound is not reasonable (although that might not happen in a realistic timeframe). There is also the question of people wanting to use non-USD currencies which will have different values.

The real question here is not what is a reasonable value for the bounds, but whether bounds are necessary at all. I.e. given that these values are ultimately based on the sum of costs of individual pieces of equipment, and those costs are in turn based on dimensions of the equipment, these values should be implicitly bounded by the dimensions of the equipment. Adding additional bounds here is essentially adding an arbitrary bound on the dimensions of the equipment that might be tighter than necessary (or alternatively is a bound that will never be active due to tighter constraints elsewhere in the model).

bpaul4:
@andrewlee94 please let me know if the changes have addressed your comments. I opened https://github.com/prommis/prommis/issues/92 and added a couple of items related to your suggestions.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -110,50 +110,55 @@ class QGESSCostingData(FlowsheetCostingBlockData):
...
     CONFIG.declare(
         "total_capital_cost",
         ConfigValue(
             default=None,
-            description="value for total capital cost; ignored if no value is "
-            "passed.",
+            description="Value for total capital cost (including equipment, "
+            "installation, and other plant costs); ignored if no value is "
+            "passed. Can be a Var, Param, or Expression with currency units, "
+            "or can specific a cost year in the cost_year argument.",
```
andrewlee94, line 135: Typo: specify

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -475,177 +473,177 @@ def build_process_costs(
...
                     bounds=(0, 1e4),
                     doc="EPCM cost in $MM",
-                    units=getattr(pyunits, "MUSD_" + CE_index_year),
+                    units=CE_index_units,
                 )
 
                 self.equipment_installation_costs = Var(
                     initialize=value(self.total_BEC),
                     bounds=(0, 1e4),
                     doc="Equipment installation EPCM cost in $MM",
```
andrewlee94, line 616: Given the change below, I do not think we should keep the "$MM" in the doc string, as it can quite easily be wrong. For determining the units of something, it is far better to directly query the units attribute than to look at a doc string that is likely static.

andrewlee94 (commented):
I found a typo and a couple of other minor things, but I think my main concerns have either been addressed here or moved to issue #92. I also see that more docs are in a different issue as well.

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: I took out all of the docstring and report references to dollars, $, or USD. The method to autodetect reference dictionary units still looks for "$", that's something that will need to be generalized when we do that currency update.

bpaul4: Also, the report method has a new line to report the currency units.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -1096,162 +1095,153 @@ def initialize_build(*args, **kwargs):
 
     def report(self, export=False):
         var_dict = {}
+        var_dict["Plant Cost Units"] = str(pyunits.get_units(self.total_plant_cost))
 
         if hasattr(self, "total_plant_cost"):
-            var_dict["Total Plant Cost [$MM]"] = value(self.total_plant_cost)
+            var_dict["Total Plant Cost"] = value(self.total_plant_cost)
```
andrewlee94, line 1101: As a passing note (not for this PR), you can use `as_quantity` (from Pyomo) to get a `pint` Quantity instead (which contains units, and has some useful features for automatically converting the value to a set of display units).

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2533,36 +2516,30 @@ def total_BEC_eq(c):
...
         if hasattr(b, "total_fixed_OM_cost"):
-            print(
-                "Total annual fixed O&M cost: $%.3f Million"
-                % value(b.total_fixed_OM_cost)
-            )
+            print("Total annual fixed O&M cost: %.3f" % value(b.total_fixed_OM_cost))
         if hasattr(b, "total_variable_OM_cost"):
             print(
-                "Total annual variable O&M cost: $%.3f Million"
+                "Total annual variable O&M cost: %.3f"
```
andrewlee94, line 2532: Another minor note, but you might want to consider `fstrings` for this (I find them easier to read).

PR #90, opened 2024-08-01
More REE Costing Improvements

bpaul4:
## Summary/Motivation:
Improves some features and adds explicit support for more imported costing types.

## Changes proposed in this PR:
- Adds support for custom costing (capital, fixed operating, variable operating)
- Extend WaterTAP support to variable operating costs, including a check for ZeroOrder models
- Fix portions of the report and variable O&M methods
- Minor cleanup and bug fixes

## Reviewer's checklist / merge requirements:
- [ ] The head branch (i.e. the "source" of the changes) is not the `main` branch on the PR author's fork
- [ ] Documentation
- [ ] Tests
- [ ] Diagnostic tests for models

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.md file
   at the top level of this directory.

2. I represent I am authorized to make the contributions and grant the license. If my employer has
   rights to intellectual property that includes these contributions, I represent that I have
   received permission to make contributions and grant the required license on behalf of that
   employer.

PR #93, opened 2024-08-14
Adds Costing for Diafiltration

mollydougher:
## Summary/Motivation:
Develop costing capabilities for membrane systems.

- Add a custom costing class for membrane systems.
- Demonstrate the cost model on the diafiltration cascade flowsheet to separate lithium and cobalt.

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+    MSContactor,
+    MSContactorInitializer,
+    Mixer,
+    MixingType,
+    MomentumMixingType,
+    MixerInitializer,
+)
+
+
+class _StateBlock(StateBlock):
```
andrewlee94, line 41: To be a little more general, it might be good to move the property package classes to a separate module. That way, they can be reused in other places if/when necessary, as well as keeping modules smaller.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
+#########################################################
```
andrewlee94, line 1: 1. There is a PrOMMiS header that should be included in every file (just copy it from another file).

2. For documentation (and pylint) purposes, it is good to have a short doc string at the start of each file describing what it does, etc. For this case, the existing comment could be part of it, but it should be put inside a block quote (`"""`).

3. You should add these modules to `/docs/api.rst` so that they appear in the autodocs.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+# MSContactor flowsheet code adapted from Andrew Lee.
+#########################################################
+
+
+import pyomo.environ as pyo
```
andrewlee94, line 6: This is generally discouraged for core code - it is better to only import the components you will actually use (this helps with maintenance too). This type of import is OK for simpler user facing examples and demos to cut down on time and explanation however.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+    UnitModelCostingBlock,
+)
+from idaes.core.util.model_statistics import degrees_of_freedom
+from idaes.core.util.initialization import fix_state_vars, propagate_state
+from prommis.nanofiltration.costing.diafiltration_cost_model import (
+    DiafiltrationCosting,
+    DiafiltrationCostingData,
+)
+
+from idaes.models.unit_models import (
```
andrewlee94, line 31: You should group imports by type, and keep them in the following order:

1. Native Python imports
2. Third part Python imports
3. Pyomo imports
4. IDAES imports
5. PrOMMiS imports.

It can also be helpful to keep each section in alphabetical order, if only to make it easier to find each import.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+    def fix_initialization_states(self):
+        """
+        Fixes state variables for state blocks.
+
+        Returns:
+            None
+        """
+        fix_state_vars(self)
+
+    def initialization_routine(self):
```
andrewlee94, line 51: This is not needed any more - this API has been deprecated and replaced by the new `Initializer` approach.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+    (
+        R_Li,
+        R_Co,
+        P_Li,
+        P_Co,
+        membrane_length,
+        feed_flow_rate,
+        diafiltrate_flow_rate,
+        permeate_flow_rate,
+    ) = get_values(m)
```
andrewlee94, line 144: An inline comment to explain what you are doing here would help. Also, a slightly more descriptive method name might also help clarify what is happening.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+        membrane_width=w,
+        water_flux=Jw,
+        feed_flow=feed_flow_rate,
+        perm_flow=permeate_flow_rate,
+        inlet_pressure=diafiltrate_inlet_pressure,
+        outlet_pressure=diafiltrate_outlet_pressure,
+        diaf_flow=diafiltrate_flow_rate,
+    )
+
+    # print(f"DOF = {degrees_of_freedom(m)}")
```
andrewlee94, line 158: Commented code like this should be cleaned up.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+        m.fs.stage1.elements,
+        m.fs.solutes,
+        rule=solute_rule,
+    )
+
+    m.fs.stage1.retentate_inlet.flow_vol[0].fix(Q_feed)
+    m.fs.stage1.retentate_inlet.conc_mass_solute[0, "Li"].fix(C_Li_feed)
+    m.fs.stage1.retentate_inlet.conc_mass_solute[0, "Co"].fix(C_Co_feed)
+
+    assert degrees_of_freedom(m.fs.stage1) == 0
```
andrewlee94, line 230: You should not have naked `asserts` like this in actual code; if it does get triggered somehow, the user will end up with a less-then-helpful error message. Either remove these checks from the actual code and save them for tests, or raise a proper exception with a useful message.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+
+    initializer = MSContactorInitializer()
```
andrewlee94, line 232: I would suggest separating the building of the model from the initialization of the model; i.e. have one function that builds the full model, and another that steps throguh initialization.

There might be rare cases you can;t make that separation, but I doubt this is one of them.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+    m.fs.stream5 = Arc(
+        source=m.fs.mix2.outlet,
+        destination=m.fs.stage3.retentate_inlet,
+    )
+    m.fs.stream6 = Arc(
+        source=m.fs.stage3.retentate_outlet,
+        destination=m.fs.mix1.inlet_1,
+    )
+
+    pyo.TransformationFactory("network.expand_arcs").apply_to(m)
```
andrewlee94, line 408: I suggest just doing one transformation of the Arcs at the end of the build method. For one, it i less chance for things to go wrong (such as getting Arcs expanded multiple times if Pyomo messes something up).

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+
+    # Unfix Feed to Mixer 1 inlet 1
```
andrewlee94, line 410: It is often also helpful to have a separate setting the variables from the build method too - for one this makes it easier for another user to take your build method but apply their own design conditions (also part of why it is helpful to separate the initialization from the build).

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+    # Lets try again
+    # Initial feed guess is pure diafiltrate (no recycle)
+    # Remember to only set the value, not fix the inlet
+    m.fs.stage3.retentate_inlet.flow_vol[0].set_value(Q_diaf)
+    m.fs.stage3.retentate_inlet.conc_mass_solute[0, "Li"].set_value(C_Li_diaf)
+    m.fs.stage3.retentate_inlet.conc_mass_solute[0, "Co"].set_value(C_Co_diaf)
+
+    initializer.initialize(m.fs.stage3)
+
+    add_expressions(m)
```
andrewlee94, line 452: A comment about what this does would be useful.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+
+    return m
+
+
+def create_rules(m):
```
andrewlee94, line 457: Not necessarily wrong, but it seems odd to me to have a function which creates the functions/rules for the `Constraints`.

Firstly, the `@model.Constraint` approach to creating constraints (and other components) is generally more compact and easier to read.
Secondly, why not just create these at the scope where they are used?

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+            m.fs.sieving_coefficient[j] - 1
+        ) * pyo.log(
+            b.retentate[0, s].flow_vol
+        )
+
+    return (solvent_rule, solute_rule, stage3_solute_rule)
+
+
+def solve_model(m):
+    solver = pyo.SolverFactory("ipopt")
```
andrewlee94, line 511: I recommend trying to use `get_solver` from IDAES if possible. For one, this means we have a standard solver across all code, and the default solver is now the new `"ipopt_v2"` interface with a linear presolve (and in future will also include model scaling by default).

Using `get_solver` means that we do not need to go change every instance of solver creation when we make changes to our solvers.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+    solver.solve(m, tee=True)
+
+
+def add_expressions(m):
```
andrewlee94, line 515: All methods/functions/classes should have a doc string explaining what it does and how it is used (including arguments and what is returned). Google style doc strings are the standard for IDAES.

src/prommis/nanofiltration/costing/diafiltration.py
```diff
@@ -0,0 +1,713 @@
...
+
+
+def add_objective(m):
+    def cost_obj(m):
+        return m.fs.costing.total_annualized_cost
+
+    m.cost_objecticve = pyo.Objective(rule=cost_obj)
+
+
+def print_information(m):
```
andrewlee94, line 675: Not critical for now, but for these types of reporting functions it is a good idea to allow the user to specify an output stream to use.

```
import sys

def my_report(model, stream=None):
    if stream is None:
        stream = sys.stdout

    stream.write("Hello world/n")
```

This is helpful for testing as it lets you provide a `StringIO` object to capture the output of the report function and check that it is correct without having to mess around with `capsys`.

andrewlee94 (commented):
I took a quick look at the `diafiltration.py` file and ended up writing far more comments than I intended.

Most are fairly simple formatting things to make the code easier to read and test.

On that note, we will need tests for all the code too.

MarcusHolly (commented):
A couple of minor comments but will review more thoroughly once this is out of draft

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -0,0 +1,408 @@
...
+            ),
+            doc="Total variable operating cost of process per operating period",
+        )
+
+        self.total_operating_cost_constraint = Constraint(
+            expr=self.total_operating_cost
+            == (self.total_fixed_operating_cost + self.total_variable_operating_cost),
+            doc="Total operating cost of process per operating period",
+        )
+        #####
```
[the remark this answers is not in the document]
mollydougher: This is a note for myself to block off a section of code. I will remove this when cleaning up my files. Thanks for pointing this out.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -397,12 +348,73 @@ def capital_cost_constraint(blk):
...
+
+        else:
+
+            @blk.Constraint()
+            def fixed_operating_cost_constraint(blk):
+                return (
+                    blk.fixed_operating_cost
+                    == 0
+                    * blk.costing_package.base_currency
+                    / blk.costing_package.base_period
```
bpaul4, line 419: In the past, Pyomo has encountered issues building expressions as 0 * [unit container]. It may be safer to remove the constraint and just fix `fixed_operating_cost` to 0 in this case. Or, set the initial value to 0 when creating the variable so we don't need the `else` case.

mollydougher: @bpaul4 @andrewlee94 I am having a DOF issue while trying to implement these changes. When I initialize the `blk.fixed_operating_cost` variable at 0 and fix it (to eliminate the need for `else`) my DOF of the cost block is now -1.

andrewlee94: You should dig into that further to see what is going on. I would start with the Diagnostics Toolbox and see what it says about structural singularities.

mollydougher: @andrewlee94 Thanks. I didn't initially see anything, and was surprised to see that initializing `blk.fixed_operating_cost` at 0, but not fixing the variable results in DOF=1. I'll dig further into the Toolbox.

mollydougher:
@adowling2 @andrewlee94 Is there any reason you can think of that since updated my environment with `requirements-dev.txt`, and not changing anything in my flowsheet, my IPOPT went from finding an optimal solution in 198 iterations, to needing 1880+ iterations and converging to a locally infeasible point?

mollydougher:
> A few, but it is hard to tell without more details. First, what Python version are you using? We just discovered a potential issue with Python 3.12. Otherwise, do you know what was changed in the update? Finally, the first guess in these cases is always scaling.

I am using Python 3.11.7. 
I thought I was just updating the version for black, but I think Pyomo may have updated to 6.8.0 (Not certain about the previous version I was using). 
I will look at the scaling to debug.

andrewlee94:
The Pyomo version could be part of the issue - did you update the IDAES version at the same time, and if so what was the old version?

I'll often only update individual dependencies to avoid some of these issues. `requirements-dev.txt` is basically a list of `pip` commands, so you can copy the appropriate line and just do `pip install [copied-text]`.

mollydougher:
> The Pyomo version could be part of the issue - did you update the IDAES version at the same time, and if so what was the old version?
> 
> I'll often only update individual dependencies to avoid some of these issues. `requirements-dev.txt` is basically a list of `pip` commands, so you can copy the appropriate line and just do `pip install [copied-text]`.

I'll have to look into that further, but noted for the future. For now, I added scaling to some of the volume flows using your workflow from the `leach_flowsheet.py` and that has improved solver performance (still takes 725 but returns my optimal solution and doesn't terminate with regularization).

mollydougher:
@andrewlee94 @bpaul4 @MarcusHolly @adowling2
This PR is ready to be officially reviewed.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -0,0 +1,406 @@
...
+            domain=NonNegativeReals,
+            doc="Maintenance-labor-chemical factor [fraction of investment cost/year]",
+            units=units.year**-1,
+        )
+        self.factor_capital_annualization = Var(
+            initialize=0.1,  # TODO: verify
+            domain=NonNegativeReals,
+            doc="Capital annualization factor [fraction of investment cost/year]",
+            units=units.year**-1,
+        )
```
MarcusHolly, line 59: Have these been verified?

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -0,0 +1,737 @@
+#####################################################################################################
+# “PrOMMiS” was produced under the DOE Process Optimization and Modeling for Minerals Sustainability
+# (“PrOMMiS”) initiative, and is copyright (c) 2023-2024 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory, et al. All rights reserved.
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license information.
+#####################################################################################################
+"""
+Diafiltration cascade flowsheet for separating lithium and cobalt. Serves as an example of
+custom cost model.
```
MarcusHolly, line 9:
```suggestion
Diafiltration cascade flowsheet for separating lithium and cobalt. Serves as an example of a
custom cost model.
```

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -0,0 +1,737 @@
...
+    def Li_recovery_constraint(m):
+        return m.Li_recovery >= 0.95
+
+    # @m.Constraint()
+    # def Li_purity_constraint(m):
+    #     return m.Li_purity >= 0.8
+
+    # @m.Constraint()
+    # def Co_purity_constraint(m):
+    #     return m.Co_purity >= 0.5
```
MarcusHolly, line 619: Is there a reason these purity constraints shouldn't be enforced?

src/prommis/nanofiltration/costing/diafiltration_cost_block.py
```diff
@@ -0,0 +1,155 @@
...
+        # self.membrane_capital_cost = Var(
+        #     initialize=0,
+        #     units=self.base_currency,
+        #     doc="Total membrane capital cost",
+        # )
+        # self.pump_capital_cost = Var(
+        #     initialize=0,
+        #     units=self.base_currency,
+        #     doc="Total pump capital cost",
+        # )
```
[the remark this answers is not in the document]
mollydougher: Good catch, I need to remove this. It's leftover from old debugging sessions.

src/prommis/nanofiltration/costing/diafiltration_cost_block.py
```diff
@@ -0,0 +1,155 @@
...
+        #     doc="Electricity cost",
+        #     units=units.USD_2018 / units.kWh,
+        # )
+        # self.defined_flows["electricity"] = self.electricity_cost
+
+        # self.electrical_carbon_intensity = Var(
+        #     initialize=0.475,
+        #     doc="Grid carbon intensity [kgCO2_eq/kWh]",
+        #     units=units.kg / units.kWh,
+        # )
```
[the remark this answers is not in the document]
mollydougher: Same as above. I'll remove this as it is no longer needed.

src/prommis/nanofiltration/costing/diafiltration_cost_block.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def _get_costing_method_for(self, unit_model):
+        # """
+        # Allow the unit model to register its default costing method,
+        # either through an attribute named "default_costing_method"
+        # or by naming the default costing method "default_costing_method"
+        # """
+        # if hasattr(unit_model, "default_costing_method"):
+        #     return unit_model.default_costing_method
+        # return super()._get_costing_method_for(unit_model)
```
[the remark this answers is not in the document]
mollydougher: Good catch, this method can be deleted as well.

src/prommis/nanofiltration/costing/diafiltration_cost_block.py
```diff
@@ -0,0 +1,155 @@
...
+        unit_model.default_costing_method()
+
+    # TODO: determine if this method is still needed
+    def register_flow_type(self, flow_type, cost):
```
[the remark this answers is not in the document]
mollydougher: It is not needed, but I had left it in the model. I will remove it.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
mollydougher: Yes, I think these are okay assumptions for now. I will remove these comments.

src/prommis/nanofiltration/diafiltration.py
mollydougher: I didn't use them for this particular example, but the user could choose to use them. I will update these to be optional constraints.

bpaul4:
@andrewlee94 I am testing @mollydougher 's code locally with the QGESS REE costing in #90 . Custom models that don't have anything fancy in the `build_global_params()` method work great when the costing block is set to the custom class, and also when the costing block is `QGESSCosting()`.

The diafiltration model works great when the costing block is set to `m.fs.costing = DiafiltrationCosting()`. However, when I set the costing block to `m.fs.costing = QGESSCosting()` there are issues due to the missing global params.

Is there a straightforward way to call `DiafiltrationCostingBlockData._build_common_global_params` and `DiafiltrationCostingData.build_global_params` to add those attributes to `m.fs.costing = QGESSCosting()`?

Another option is to handle models with custom classes the same way as WaterTAP models that have their own costing classes. Users would need to specify the class they are using for a certain model, and the REE costing will search for variables named `capital_cost`, `fixed_operating_cost`, and `variable_operating_cost` to add to the REE cost summations.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -0,0 +1,406 @@
...
+)
+
+
+@declare_process_block_class("DiafiltrationCosting")
+class DiafiltrationCostingData(DiafiltrationCostingBlockData):
+    """
+    Costing block for the diafiltration flowsheet
+    """
+
+    def build_global_params(self):
```
bpaul4, line 28: @mollydougher For compatibility with the `QGESSCosting` class in `ree_plant_capcost`, the global parameters should only contain the base currency and year. This way, a flowsheet will set the base currency and year based on whether the `DiafiltrationCosting` or `QGESSCosting` class is used and everything else is contained within the costing model itself.

The `QGESSCosting` class expects to find capital costs (bare equipment only, not including installation and not adjusted or annualized) as well as fixed costs related to initial construction and variable costs related to unit-specific resource usage. I'd suggest moving the `_build_common_global_params()` call inside your `build_process_costs` method as they are specific to your module. Similarly, `factor_total_investment`, `factor_maintenance_labor_chemical`, `factor_capital_annualization`, and `capital_recovery_factor` should move inside the `build_process_costs` method. It might make sense to move `density` and `specific_gravity` inside `cost_pump` if that is the only place they are used. If `electricity_cost` is used in only a couple places, it's probably okay to redefine it.

One last note, I don't see where `variable_operating_cost` is being calculated, although it appears in the aggregate cost section. Costs from resources (electricity, chemicals, water, waste streams with a known disposal cost, ...) should be included in variable OPEX, while costs from any materials or other items related to construction including membrane replacement can be fixed OPEX. The QGESS class already includes estimates of maintenance, labor, admin, taxes, etc., as well as installation costs for equipment.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -0,0 +1,406 @@
...
+                    * blk.membrane_cost
+                    * blk.membrane_area
+                ),
+                to_units=blk.costing_package.base_currency
+                / blk.costing_package.base_period,
+            ) + units.convert(
+                (blk.SEC * vol_flow_perm * blk.costing_package.electricity_cost),
+                to_units=blk.costing_package.base_currency
+                / blk.costing_package.base_period,
+            )
```
bpaul4, line 281: This second term seems more like a variable operating cost to me.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def cost_pump(blk, inlet_pressure, outlet_pressure, inlet_vol_flow):
```
bpaul4, line 283: Same note on default values as above.

bpaul4 (commented):
Some more comments.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -0,0 +1,406 @@
...
+            self.total_operating_cost, self.total_operating_cost_constraint
+        )
+
+    def cost_membranes(
+        blk,
+        membrane_length,
+        membrane_width,
+        water_flux,
+        vol_flow_feed,
+        vol_flow_perm,
```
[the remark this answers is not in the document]
andrewlee94: @bpaul4 Is it necessarily "safer"? By not providing defaults, it forces the user to provide values which is not always a bad thing (defaults can lead to lazy users and this failures due to things they overlooked).

Either way, it is more a matter of ensuring that there is necessary error handling and how that fits into the API.

src/prommis/nanofiltration/costing/diafiltration_cost_block.py
```diff
@@ -0,0 +1,85 @@
...
+
+from idaes.core import FlowsheetCostingBlockData, declare_process_block_class
+
+
+@declare_process_block_class("DiafiltrationCostingBlock")
+class DiafiltrationCostingBlockData(FlowsheetCostingBlockData):
+    """
+    The base class for costing diafiltration-precipitation flowsheets
+
+    methods that come directly from reference file:
```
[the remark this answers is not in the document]
mollydougher: This was a note for myself that can be removed.

src/prommis/nanofiltration/costing/diafiltration_cost_block.py
```diff
@@ -0,0 +1,85 @@
...
+            initialize=3.4,  # TODO: verify
+            doc="Total Purchased Equipment Cost (TPEC)",
+            units=units.dimensionless,
+        )
+
+        self.TIC = Var(
+            initialize=1.65,  # TODO: verify
+            doc="Total Installed Cost (TIC)",
+            units=units.dimensionless,
+        )
```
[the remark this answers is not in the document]
mollydougher: They are defined in the docstrings (L75 and L81). These are the variable names that the WaterTAP costing package uses - would you recommend updating the variable names to something different to comply with the UKy costing?

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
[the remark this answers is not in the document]
mollydougher: Instead of providing defaults, should I write Exceptions for if the user does not specify these arguments?

src/prommis/nanofiltration/costing/diafiltration_cost_block.py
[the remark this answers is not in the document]
bpaul4: I read too quickly and missed the docstrings - this is good as is.

src/prommis/nanofiltration/tests/test_diafiltration.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+from pyomo.environ import value
+
+from idaes.core.util.model_diagnostics import DiagnosticsToolbox
+from idaes.models.unit_models import MSContactor
+
+import pytest
+
+pytest.importorskip("watertap", reason="WaterTAP dependency not available")
```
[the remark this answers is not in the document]
mollydougher: I think I don't need this anymore. Per Marcus' feedback, I removed the last remaining WaterTAP import left in the model this morning. Thanks for catching this.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
mollydougher: Thanks for this feedback, @bpaul4. Everything makes sense. After moving around where those variables are defined, I ran into some DOF issues that I'm still debugging. I should have a commit up sometime tomorrow to address these and your remaining comments.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -29,72 +37,47 @@ def build_global_params(self):
...
         # Set the base year for all costs
         self.base_currency = units.USD_2021
 
         # Set a base period for all operating costs
         self.base_period = units.year
 
-        # the following global parameters are from the reference file
-        self.factor_total_investment = Var(
+        # fix the parameters
+        self.fix_all_vars()
```
bpaul4, line 47: What does this call do?

mollydougher: Actually, this is not needed, since I removed all other global parameter declarations in this method.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
```diff
@@ -222,6 +190,69 @@ def membrane_area_equation(blk):
...
+                (
+                    blk.factor_membrane_replacement
+                    * blk.membrane_cost
+                    * blk.membrane_area
+                ),
+                to_units=blk.costing_package.base_currency
+                / blk.costing_package.base_period,
+            )
+
+    def cost_membrane_pressure_drop(
```
bpaul4, line 209: It looks like the membrane and cascade "shell" are separate units now, where the membrane has material costs (fixed OPEX) and no power usage, and the cascade has power usage (variable OPEX) but no material costs. Am I understanding that correctly?

bpaul4: Also, should the cascade have an associated CAPEX?

mollydougher: Yes. After updating how the global parameters were defined, I removed the arbitrary `UnitModelBlock` of `m.fs.membrane`, and instead created `UnitModelCostingBlock`s for each membrane stage. However, the membrane pressure drop calculation (now attached to `m.fs.cascade.costing`) is really intended to consider the overall system. This pressure drop is very small for stage 1 and 2. 

And yes, I believe we will want to include material cost for cascade construction. I have to investigate whether a standard method for this exists.

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -134,12 +125,71 @@ def build_model():
...
 
 
+def add_global_flowsheet_parameters(m):
+    """
+    Adds global parameters to the Pyomo model
+
+    Args:
+        m: Pyomo model
+    """
+    m.Jw = Param(
```
bpaul4, line 141: This is nice @mollydougher, thanks for adding this. @andrewlee94 should the code specify that is `Param` has `mutable=True`, or is this OK?

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -521,7 +575,7 @@ def solve_model(m):
     # TODO: add Boolean variable to calculate pump OPEX
     # Verify the feed pump operating pressure workaround is valid
     # assume this additional cost is less than half a cent
-    if value(m.fs.feed_pump.costing.fixed_operating_cost) >= 0.005:
+    if value(m.fs.feed_pump.costing.variable_operating_cost) >= 0.005:
         raise ValueError(
             "The fixed operating cost of the feed pump as calculated in the feed"
```
bpaul4, line 580:
```suggestion
            "The variable operating cost of the feed pump as calculated in the feed"
```

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -615,7 +693,7 @@ def Li_recovery_constraint(m):
 
         @m.Constraint()
         def Li_purity_constraint(m):
-            return m.Li_purity >= 0.8
+            return m.Li_purity >= 0.5
```
bpaul4, line 696: Why did this change?

mollydougher: I was adjusting these values when I was debugging the scaling/solver convergence. These constraints are not currently active for this example, though a user could choose to change this if desired.

src/prommis/nanofiltration/diafiltration.py
mollydougher: Good catch, thanks!

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
bpaul4: If it's a vessel shell, you can see if the IDAES reactor vessel method is appropriate for the size and shape of your cascade: https://github.com/IDAES/idaes-pse/blob/fe422c006a4b2e03b9b2b1a6ba39eba0043380d1/idaes/models/costing/SSLW.py#L431

src/prommis/nanofiltration/diafiltration.py
andrewlee94: If a `Param` has units, it is `always `mutable` and will log warnings if this is not set explicitly, so in this case the `Param` should be mutable.

More generally, you should decide if it makes sense for a parameter to change. E.g. acceleration due to gravity is unlikely to change, but a flux parameter like this would change depending on the membrane used and a number of other factors.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
mollydougher: Actually, after looking into this more, it seems like the cost/area of the membranes (accounted for in the capital cost of the stages) includes the whole spiral wound membrane module. Typically, I believe the membrane cascades are "open" and there wouldn't be one large vessel shell around all of the modules. I think we can omit `m.fs.cascade.costing.capital_cost` at least for now.

src/prommis/nanofiltration/costing/diafiltration_cost_model.py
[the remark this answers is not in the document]
bpaul4: Yes, thank you!

src/prommis/nanofiltration/diafiltration.py
bpaul4: Thanks! Could you please add tests to cover those exceptions? In the File Comparison, CodeCov flags uncovered lines.

src/prommis/nanofiltration/tests/test_diafiltration.py
```diff
@@ -0,0 +1,119 @@
...
+    main,
+)
+
+
+@pytest.mark.component
+def test_main():
+    """
+    Tests the execution of the main function in diafiltration.py
+    """
+    m = main()
```
bpaul4, line 31: Would it be reasonable to move some of the insanitation outside of your methods? Specifically so that your methods don't create/overwrite `m`, `m.fs`, and `m.fs.costing` internally. It would greatly help integrating your model into the QGESS costing if the following workflow could be run:

```
m = ConcreteModel()
m.fs = FlowsheetBlock(dynamic=False)
m.fs.costing = CouldBeAnyCostingBlock()  # for example, QGESSCosting() or DiafiltrationCosting()

# could build other blocks in the flowsheet, including other UnitModelCostingBlocks

# Then, build the DF model and costing using a single method call
from prommis.nanofiltration import main  # method to build entire model, costing, and optimization equations
main(m)  # expects m, m.fs, and m.fs.costing to already exist - also allows users to define these themselves
```

andrewlee94: @bpaul4 I am not quite sure what your concern is here. It looks like each test is creating a new instance of the model, so nothing should be overwritten. 

Are you suggesting the use of a `pytest.fixture` so that `main` is only run once and used for all the tests? Whilst this does reduce the amount of time spent construction models, you do need to be careful that tests do not pollute each other.

One thing I will say however - I do not think `main` should be imported from another module. If you really need this, then it probably should have a better name (such as `build_model`); however I note you already imported `build_model` which makes me wonder if `main` is really needed.

bpaul4: I'm thinking more about using this model as part of a larger flowsheet where `m`, `m.fs,` and `m.fs.costing` already exist. As they are now, these methods recreate those objects.

bpaul4: Also, `m.fs.costing` is defined as `DiafiltrationCosting` when `add_costing` runs, so using `QGESSCosting` means duplicating all of the costing code rather than just calling the costing method.

andrewlee94: For this particular file, I don't see that as a problem as this is a test file for the main code (which might be part of my confusion). It might be more useful to comment on specific parts of the actual methods on how to address any limitations that might exist.

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -0,0 +1,774 @@
...
+
+
+def build_model():
+    """
+    Builds the diafiltration flowsheet
+
+    Returns:
+        m: Pyomo model
+    """
+    m = ConcreteModel()
```
bpaul4, line 111: Remove this line so that `m` refers to an existing model object. The method would take `m` as an input, like `def build_model(m)`.

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -0,0 +1,774 @@
...
+    add_global_flowsheet_parameters(m)
+
+    m.fs = FlowsheetBlock(dynamic=False)
```
bpaul4, line 114: Same here, remove this line so that it can use an existing flowsheet block.

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -0,0 +1,774 @@
...
+
+    Args:
+        m: Pyomo model
+    """
+    # creating dummy variables to store the UnitModelCostingBlocks
+    m.fs.cascade = UnitModelBlock()  # to cost the pressure drop
+    m.fs.feed_pump = UnitModelBlock()  # to cost feed pump
+    m.fs.diafiltrate_pump = UnitModelBlock()  # to cost diafiltrate pump
+
+    m.fs.costing = DiafiltrationCosting()
```
bpaul4, line 609: Same here, remove this line so that users can define their own flowsheet-level costing block.

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -0,0 +1,774 @@
...
+
+
+def main():
+    """
+    Builds and solves the diafiltration flowsheet with cost
+
+    Returns:
+        m: Pyomo model
+    """
+    m = build_model()
```
bpaul4, line 63: This would be `def main(m)` and `build_model(m)`

src/prommis/nanofiltration/diafiltration.py
```diff
@@ -0,0 +1,774 @@
...
+
+    print("\ntotal membrane area")
+    print(f"{value(m.membrane_length)*value(m.w)} m2")
+
+    print("\ntotal annualized cost")
+    print(f"${value(m.fs.costing.total_annualized_cost)}")
+
+
+if __name__ == "__main__":
+    main()
```
bpaul4, line 774: This would become
```
m = ConcreteModel()
m.fs = FlowsheetBlock(dynamic=False)
m.fs.costing = DiafiltrationCosting()
main(m)
```

The test file would do something similar.

bpaul4 (commented):
I added some comments to clarify my request above.

src/prommis/nanofiltration/diafiltration.py
andrewlee94: I do not think `main` should take any arguments, as that defeats the purpose of a `main()` method.

src/prommis/nanofiltration/diafiltration.py
andrewlee94: @mollydougher What is the intended use case of these methods? Is the purpose to build a specific case study, or are they intended to be more general for building a generic diafiltration cascade?

andrewlee94: If that is the case, then I think this is OK as is. However, I suggest making it clear in a doc string that this is a specific case study, and maybe add some inline comments to help readers understand what is happening (and to teach them so they can crate their own implementations later).

src/prommis/nanofiltration/diafiltration.py
mollydougher: @andrewlee94 This makes sense and I removed the import of `main` in `test_diafiltration.py`. Now the Codecov check is failing due to uncovered lines in `main()` - I am not sure how to cover those lines if `main()` is no longer being tested.

src/prommis/nanofiltration/diafiltration.py
mollydougher: @bpaul4 Please let me know if this makes sense to you. Otherwise, I am not sure how to reconcile your comments without giving arguments to `main()`.

src/prommis/nanofiltration/diafiltration.py
andrewlee94: Given this is a test for the case study, you should import and test `main` here. The point of the previous discussion is that `main` should not be viewed as something that other code should important use (and thus it does not need to take an existing model as an argument).

mollydougher: Noted, thanks! I'll update this.

mollydougher:
@bpaul4 @andrewlee94 @MarcusHolly @adowling2 I believe all outstanding comments have been addressed.

andrewlee94 (approved):
Looks good to me.

PR #94, opened 2024-08-15
Cost Estimation of a Hydrogen Decrepitation Furnace

Akintomiwaojo:
## Summary/Motivation:
This work estimates the cost of a hydrogen decrepitation furnace using a bottom-up approach.

-Add a file for estimating the cost of a hydrogen decrepitation furnace. 
-Add a test file.

## Reviewer's checklist/merge requirements:
- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/uky/costing/models/hd_ce.py
```diff
@@ -0,0 +1,1053 @@
...
+)
+from idaes.core.util.constants import Constants
+from idaes.core.util.math import smooth_max
+import idaes.logger as idaeslog
+
+_log = idaeslog.getLogger(__name__)
+
+
+def custom_REE_plant_currency_units():
+    """
```
MarcusHolly, line 23: I might consider renaming this file to hydrogen_decrepitation to make it more clear for users

MarcusHolly (commented):
Some minor comments

src/prommis/uky/costing/models/hd_ce.py
```diff
@@ -0,0 +1,1053 @@
...
+        thermal_cond_metal_material2=45,  # W/(m*K)
+        hours_per_shift=8,
+        shifts_per_day=3,
+        specific_heat_capacity_insulation1=1.08,  # KJ/(kg*K)
+        specific_heat_capacity_metal1=0.468,  # KJ/(kg*K)
+        specific_heat_capacity_insulation2=0.9,  # KJ/(kg*K)
+        specific_heat_capacity_metal2=0.502416,  # KJ/(kg*K)
+        operating_days_per_year=336,
+        efficiency=0.95,
+        electricity_rate=0.081,  # USD/Kwhr
```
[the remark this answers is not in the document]
andrewlee94: A note here - it might be better to explicitly include units of measurement here for verification and conversion purposes rather than have an implicit assumption. If you do it right, this means that the value users provide will be automatically converted to the desired units, no matter what units they use.

I.e. `electricity_rate=0.081*units.USDXXXX/units.kW/units.hour`

src/prommis/uky/costing/models/tests/test_hd_ce.py
```diff
@@ -0,0 +1,807 @@
...
+                "operating_days_per_year": 336,  # days
+                "efficiency": 0.95,
+                "electricity_rate": 0.081,  # USD/kWhr
+                "labor_rate": 75,  # USD/hr
+                "temperature_controller_price": 129.00,  # USD
+                "number_of_units": 1,
+                "engineering_and_drafting": 1000,  # USD
+            },
+        )
+
```
bpaul4, line 88: Add checks on `model.fs.hydrogen_decrepitation_furnace.costing.costing_package.base_currency` and `model.fs.hydrogen_decrepitation_furnace.costing.costing_package.base_period`. Since you set `m.fs.costing = REEEquipmentCosting()`, we expect the units to be `units.USD_Jan_2024` and `units.year`, respectively.

src/prommis/uky/costing/models/tests/test_hd_ce.py
```diff
@@ -0,0 +1,807 @@
...
+    def test_solve1(self, model):
+        results = solver.solve(model, tee=True)
+        assert_optimal_termination(results)
+
+    @pytest.mark.unit
+    def test_solved_model_diagnostics(self, model):
+        dt = DiagnosticsToolbox(model)
+        dt.assert_no_numerical_warnings()
+
+    @pytest.mark.component
```
bpaul4, line 388: Probably need to update these values since they will be reported in `USD_Jan_2024` now.

src/prommis/uky/costing/models/tests/test_hd_ce.py
```diff
@@ -0,0 +1,807 @@
...
+                "specific_heat_capacity_metal2": 0.502416,  # KJ/(kg*K)
+                "operating_days_per_year": 336,  # days
+                "efficiency": 0.95,
+                "electricity_rate": 0.081,  # USD/kWhr
+                "labor_rate": 75,  # USD/hr
+                "temperature_controller_price": 129.00,  # USD
+                "number_of_units": 1,
+                "engineering_and_drafting": 1000,  # USD
+            },
+        )
```
bpaul4, line 473: Add base currency and period checks here too.

src/prommis/uky/costing/models/tests/test_hd_ce.py
```diff
@@ -0,0 +1,807 @@
...
+    def test_solve2(self, model):
+        results = solver.solve(model, tee=True)
+        assert_optimal_termination(results)
+
+    @pytest.mark.unit
+    def test_solved_model_diagnostics2(self, model):
+        dt = DiagnosticsToolbox(model)
+        dt.assert_no_numerical_warnings()
+
+    @pytest.mark.component
```
bpaul4, line 773: Same as above, might need to fix these results.

bpaul4 (commented):
Thanks @Akintomiwaojo for developing this model! I left some comments below.

src/prommis/uky/costing/models/hd_ce.py
```diff
@@ -0,0 +1,1053 @@
...
+            initialize=weight_insulation1,
+            units=units.kg,
+            mutable=True,
+        )
+
+        @blk.Expression(doc="Total weight of insulation material 1")
+        def total_weight_insulation1(blk):
+            return blk.quantity_insulation1 * blk.weight_insulation1
+
+        blk.price_insulation1 = Param(
```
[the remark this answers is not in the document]
andrewlee94: @bpaul4 You do not need the `units.convert` - anything going the `var.set_value()` automatically goes through unit conversion to the units of `var`.

src/prommis/uky/costing/models/hd_ce.py
```diff
@@ -0,0 +1,1053 @@
...
+                + blk.labor_cost
+                + blk.engineering_and_drafting
+                + blk.overhead_cost
+            )
+
+        @blk.Expression(doc="Base cost for all installed units")
+        def base_cost(blk):
+            return blk.base_cost_per_unit * blk.number_of_units
+
+        @blk.Expression(doc="capital expenditure in USD_Jan_2024")
```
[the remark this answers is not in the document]
andrewlee94: Also, all doc strings should start with capital letters.

src/prommis/uky/costing/models/hydrogen_decrepitation.py
```diff
@@ -174,29 +174,24 @@ def cost_hydrogen_decrepitation_furnace(
...
-            initialize=sample_volume, units=units.m**3, within=Any, mutable=True
-        )
-        blk.sample_density = Param(
-            initialize=sample_density, units=units.kg / (units.m**3), mutable=True
-        )
-        blk.chamber_to_sample_ratio = Param(
-            initialize=chamber_to_sample_ratio, units=units.dimensionless, mutable=True
-        )
+        blk.sample_volume = Param(units=units.m**3, within=Any, mutable=True)
+        blk.sample_volume.set_value(sample_volume)
```
bpaul4, line 178: What happens if `sample_mass` is not set and `sample_volume` is also not set? I don't think you test this case, and it looks like that would fail.

bpaul4: I suggest adding a check before `sample_mass` and `sample_volume` are used that does

`if sample_volume is None and sample_mass is None:`
    `[a graceful exception that fails with an error message]`

Then, add a test that fails on purpose to test the exception. Check the REE costing test cost for how to do this, there are lots of tests that have `with pytest.raises`.

bpaul4 (approved):

MarcusHolly (approved):
LGTM

andrewlee94 (changes requested):
Firstly, I do not think this code is going to the right place.

Currently, this is committing the code in the UKy flowsheet folder, but hydrogen decrepitaiton is not applicable to that case. This code should probably be committed to a new folder at the `/prommis` level.

src/prommis/uky/costing/models/hydrogen_decrepitation.py
andrewlee94: @Akintomiwaojo As @bpaul4 indicated, if you have cases like this where it is required for the user to provide certain information, then you need to have some sort of check in the code to catch cases where that is not the case and return a useful error message to the user. Otherwise, someone will try to provide different information (or insufficient information) and hte model will fail to solve but the user will have no idea of why.

bpaul4:
> Firstly, I do not think this code is going to the right place.
> 
> Currently, this is committing the code in the UKy flowsheet folder, but hydrogen decrepitaiton is not applicable to that case. This code should probably be committed to a new folder at the `/prommis` level.

Would it make sense to move this to `prommis/src/prommis/hydrogen_decrepitation/costing/`?

andrewlee94:
@bpaul4 That would make sense, although I will ask if the `costing` sub-folder is necessary (i.e. why not just `prommis/src/prommis/hydrogen_decrepitation`).

On that note however, where is the related unit model? It seems odd to have a costing module but no unit model to go with it.

lbianchi-lbl:
@bpaul4 Based on what I understand from @Akintomiwaojo's summary during today's dev call, I'm converting this back to "Draft" while we wait for the unit model to be available.

bpaul4:
The unit model described in #98 will be developed in the near future as part of ongoing milestones. Once that is merged, this pull request will be updated with appropriate tests and re-reviewed.

lbianchi-lbl:
This needs to wait for the PR corresponding to #98 that @bpaul4 will open this week. This is still on track for the June release.

bpaul4:
This depends on #155, I'm attempting to change the head to allow the costing tests to run.

PR #97, opened 2024-09-11
NPV Loan Repayment

bpaul4:
## Summary/Motivation:
Accounts for cash flows due to loan repayment and interest owed.

- Add loan repayment to NPV formula
- Fixes related to capital expenditure period

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


Akintomiwaojo (commented):
On line 5407, why do we have a flat fee of $100,000 in the royalty formula expression, but I'm not sure it was added to the expression for pv_royalties in the main ree costing file?

bpaul4:
> On line 5407, why do we have a flat fee of $100,000 in the royalty formula expression, but I'm not sure it was added to the expression for pv_royalties in the main ree costing file?

This is an example of using a custom expression to calculate the royalties, so I used arbitrary numbers.

Akintomiwaojo:
> > On line 5407, why do we have a flat fee of $100,000 in the royalty formula expression, but I'm not sure it was added to the expression for pv_royalties in the main ree costing file?
> 
> This is an example of using a custom expression to calculate the royalties, so I used arbitrary numbers.

So it's similar to the debt formula expression, right? I can see that it's a sum of fractions of the CAPEX and OPEX; however, in the costing file, loan debt is only a fraction of the CAPEX.

bpaul4:
> > > On line 5407, why do we have a flat fee of $100,000 in the royalty formula expression, but I'm not sure it was added to the expression for pv_royalties in the main ree costing file?
> > 
> > 
> > This is an example of using a custom expression to calculate the royalties, so I used arbitrary numbers.
> 
> So it's similar to the debt formula expression, right? I can see that it's a sum of fractions of the CAPEX and OPEX; however, in the costing file, loan debt is only a fraction of the CAPEX.

Yes, you're correct. The default is to calculate royalties as a percentage of revenue and debt as a percentage of CAPEX, however one or both may be defined using custom expressions. That's what the last few tests in the test file are checking, and the choice of expressions is arbitrary just to test the functionality.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2862,17 +2898,28 @@ def calculate_NPV(b, fixed_OM, variable_OM):
...
         equipment purchasing, gCap is the capital escalation growth rate
-        expressed as a decimal, and iLoan_% is the capital equipment loan
-        interest rate expressed as a decimal. The capital costs spent in each
-        year are handled separately to properly account for the value growth
-        over time.
+        expressed as a decimal. The capital costs spent in each year are handled
+        separately to properly account for the value growth over time. Loan
+        repayment and interest owed are calculated as
+
+        Annual_Loan_Payment = Debt * A/P(iLoan_%, 0, Nloan) = Debt / P/A(iLoan_%, 0, Nloan)
```
Agfritz, line 2912: Would it make sense to combine the annual loan payment and interest into a single term for simplicity?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2862,17 +2898,28 @@ def calculate_NPV(b, fixed_OM, variable_OM):
...
+
+        PV_Loan_Interest_Owed = P/A(r, iLoan_%, Nloan) * Annual_Loan_Payment - Debt
+
+        where Annual_Loan_Payment is the required combined principal and interest
+        that should be paid annually to pay off the loan on-time, Debt is the
+        loan principal (typically a percentage of the CAPEX), Nloan is the loan
+        repayment period, and iLoan_% is the capital equipment loan interest rate
+        expressed as a decimal. Note that while the annual payment has growth of
+        0 and maintains a constant present value over time, the loan balance does
```
Agfritz, line 2921: It may be helpful to clarify that the loan balance would not grow (since the balance is declining with payment), but the present value of all loan payments would grow.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -3084,27 +3182,114 @@ def pv_capital_cost_constraint(c):
...
+                    to_units=c.cost_units,
+                )
+
+        else:
+
+            b.loan_debt = Reference(b.config.debt_expression)
+
+        @b.Constraint()
+        def loan_annual_payment_constraint(c):
+            # Annual Loan Payment = Debt * A/P(%interest, 0, loan_length) = Debt / P/A(%interest, 0, loan_length)
```
Agfritz, line 3213: This conversion is from present value to future annual payments. However, I believe we would need to convert this back in order to get the present value of annual payments.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -3084,27 +3182,114 @@ def pv_capital_cost_constraint(c):
...
-                        0,
-                    ),  # formula gives P/A (r, g, 0) = 1
+                        0,  # payments adjusted to present value are constant
+                        c.capital_loan_repayment_period / pyunits.year,
+                    ),
+                    to_units=c.cost_units,
+                )
+
+        @b.Constraint()
+        def pv_loan_interest_constraint(c):
```
Agfritz, line 3247: Ultimately I think the present value of the loan payments + interest should be the following:
series present worth factor (same as used throughout)* interest rate* loan principal/ (1- (1+ interest rate)^(-n)). Does this come out to be the same thing?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -208,6 +208,32 @@ class QGESSCostingData(FlowsheetCostingBlockData):
...
+            "debt_expression is not None. The value should be a percentage, "
+            "for example 10 for a debt corresponding to 10% of the CAPEX. Set "
+            "to zero to indicate no loans are taken out on capital.",
+        ),
+    )
+    CONFIG.declare(
+        "debt_expression",
+        ConfigValue(
+            default=None,
+            description="Set the value or expression to calculate total debt.",
```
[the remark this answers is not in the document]
bpaul4: This would be a Pyomo Expression using flowsheet variables. I'll add a note that the units should be in flowsheet dollars. It's the user's choice, this could be an Expression using any variables - e.g. a percentage of the total CAPEX + a different percentage of the total OPEX, a percentage of the total CAPEX for a subset of equipment accounts, or any other custom expression the user wants to define.

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: Since we're using the series present worth factor now, I agree that it would make sense to combine and simplify the expressions.

src/prommis/uky/costing/ree_plant_capcost.py
bpaul4: Our two formulations are similar. Mine simplifies to

$\frac{P_A (r, i, N)}{P_A (i, 0, N)} * Debt$

while yours simplifies to

$\frac{P_A (r, i, N)}{P_A (i, 0, N)} * Debt$

where $P_A$ is the series present worth factor, $r$ is the discount rate, $i$ is the interest rate, $N$ is the loan length, and $Debt$ is the loan principal. So I have the annual payment series growing over time by the interest, and in yours it doesn't grow over time.

Agfritz: Ok thank you brandon for simplifying the expression! I am not familiar with including escalation/ growth for a debt payment (as long as the interest rate stays constant).

bpaul4: I thought about this some more, and I think you're correct that the annual payments shouldn't be escalating over time. There are two possible formulas we could use to get the total sum of annual payments over the repayment period:

1. Your suggestion, $TotalLoanPlusInterestPayments = \frac{P_A (r, 0, N)}{P_A (i, 0, N)} * Debt = P_A (r, 0, N) * Annual_Payment_Amount$ which assumes that the calculated annual payments are absolute amounts that must be adjusted to present day dollars. The problem here is that when the interest rate is less than the discount rate, we get a positive present value - so we're making money by taking out the loan. When using the WACC as the discount rate for a 10 year loan, the total interest is only 1% of the principal, which is very small compared to the interest rate of 6%.

2. $TotalLoanPlusInterestPayments = \frac{N}{P_A (i, 0, N)} * Debt = N * Annual_Payment_Amount$ which assumes that the calculated annual payments are already in present day dollars are we can just multiply by the number of payments. This will always be greater than the original Debt, so there is always going to be a negative present value - a cost of taking out the loan. For a 10 year loan at 6% interest, the total interest is 35% of the principal, which seems more reasonable to me since the unpaid debt balance is accumulating more interest each year.

Agfritz: Hi Brandon- thanks so much for looking into this further. I agree with you that the second approach makes more sense. I am good with going ahead with that!

bpaul4: I implemented the formulation discussed in recent meetings and offline and consolidated into a single expression.

Akintomiwaojo (approved):
LGTM!

Agfritz (approved):
The formula changes look good to go!

PR #101, opened 2024-11-01
Economy of numbers

Akintomiwaojo:
## Summary/Motivation:
Adds Economy of Number, a scaling function that estimates the future profitability of novel/First-Of-A-Kind (FOAK) equipment.

-
-

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.

### DIFF
--- a/src/prommis/uky/costing/ree_plant_capcost.py
+++ b/src/prommis/uky/costing/ree_plant_capcost.py
@@ -3460,17 +3460,21 @@ class QGESSCostingData(FlowsheetCostingBlockData):
         X is the cumulative number of units, b is the learning rate exponent, and R is the learning
         rate constant.
 
+        The equations above are derived from Faber G, Ruttinger A, Strunge T, Langhorst T, Zimmermann A,
+        van der Hulst M, Bensebaa F, Moni S and Tao L (2022) Adapting Technology Learning Curves for
+        Prospective Techno-Economic and Life Cycle Assessments of Emerging Carbon Capture and Utilization
+        Pathways. Front. Clim. 4:820261. doi: 10.3389/fclim.2022.820261
+
         Args:
             cum_num_units: The cumulative number of units.
             cost_FOAK: The cost of manufacturing the First-of-A-Kind equipment.
             CE_index_year: year for cost basis, e.g., "2021" to use 2021 dollars
             learning_rate: ranges between 0.01 - 0.1, depending on the level of maturity
                            (i.e., experimental, growing, proven, etc.)
-        Ref.:
-            Rubin, E. S., Mantripragada, H., and Zhai, H.,
-            "An Assessment of the NETL Cost Estimation Methodology".
-            Department of Engineering and Public Policy, Carnegie Mellon University,
-            Pittsburgh, PA (2016). p. 31, Fig. 6-4.
+                            as described in Rubin, E. S., Mantripragada, H., and Zhai, H.,
+                            "An Assessment of the NETL Cost Estimation Methodology".
+                            Department of Engineering and Public Policy, Carnegie Mellon University,
+                            Pittsburgh, PA (2016). p. 31, Fig. 6-4.
         """
 
         blk.cum_num_units = Param(
--- a/src/prommis/uky/costing/ree_plant_capcost.py
+++ b/src/prommis/uky/costing/ree_plant_capcost.py
@@ -3484,12 +3484,6 @@ class QGESSCostingData(FlowsheetCostingBlockData):
             units=pyunits.dimensionless,
             doc="The learning factor reflects the level of maturity of the unit/technology",
         )
-        blk.cost_FOAK = Param(
-            initialize=cost_FOAK,
-            mutable=True,
-            doc="Cost of the First-of-A-Kind of the unit",
-            units=getattr(pyunits, "MUSD_" + CE_index_year),
-        )
 
         blk.cost_NOAK = Var(
             initialize=1e5,
@@ -3507,6 +3501,7 @@ class QGESSCostingData(FlowsheetCostingBlockData):
 
         @blk.Constraint()
         def cost_NOAK_eq(b):
-            return b.cost_NOAK == b.cost_FOAK * (
-                (b.cum_num_units) ** -(b.learning_rate_exponent)
+            return b.cost_NOAK == pyunits.convert(
+                cost_FOAK * ((b.cum_num_units) ** -(b.learning_rate_exponent)),
+                to_units=getattr(pyunits, "MUSD_" + CE_index_year),
             )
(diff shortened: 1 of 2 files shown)


src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -3442,3 +3443,70 @@ def assert_Pyomo_object(b, name):
...
+            mutable=True,
+            units=pyunits.dimensionless,
+            doc="The learning factor reflects the level of maturity of the unit/technology",
+        )
+        blk.cost_FOAK = Param(
+            initialize=cost_FOAK,
+            mutable=True,
+            doc="Cost of the First-of-A-Kind of the unit",
+            units=getattr(pyunits, "MUSD_" + CE_index_year),
+        )
```
bpaul4, line 3492: Instead of making this a parameter, why not just use `cost_FOAK` directly? This will also ensure the model is dynamic, and will update/resolve if things change upstream.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -3442,3 +3443,70 @@ def assert_Pyomo_object(b, name):
...
+        def learning_rate_exponent(b):
+
+            return -log10(1 - b.learning_rate) / log10(2)
+
+        @blk.Constraint()
+        def cost_NOAK_eq(b):
+            return b.cost_NOAK == b.cost_FOAK * (
+                (b.cum_num_units) ** -(b.learning_rate_exponent)
+            )
```
bpaul4, line 3512: Wrap the right hand side in a units conversion, this will catch any case where `cost_FOAK` doesn't have currency units:

```suggestion
        @blk.Constraint()
        def cost_NOAK_eq(b):
            return b.cost_NOAK == pyunits.convert(
                b.cost_FOAK * (
                    (b.cum_num_units) ** -(b.learning_rate_exponent)
                ),
                to_units=getattr(pyunits, "MUSD_" + CE_index_year)
            )
```

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -6314,3 +6314,32 @@ def test_REE_costing_royalty_expression_solve():
...
+
+    dt = DiagnosticsToolbox(model=m, variable_bounds_violation_tolerance=1e-4)
+    dt.assert_no_structural_warnings()
+
+    solver = get_solver()
+    results = solver.solve(m, tee=True)
+    assert_optimal_termination(results)
+    dt.assert_no_numerical_warnings()
+
+    assert pyo.value(m.fs.costing.cost_NOAK[None]) == pytest.approx(36.574, rel=1e-4)
```
bpaul4, line 6345: Why is `cost_NOAK` indexed here?

Akintomiwaojo: I used `None` similarly to how it was used for pv_royalties, like accessing a scalar value of an expression. But looking closer now, I can see I don't need to index `cost_NOAK` as it is a variable, not an expression. Thanks for pointing this out.

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -3442,3 +3443,65 @@ def assert_Pyomo_object(b, name):
                     f"Argument {obj} of type {type(obj)} is not a supported object type. "
                     f"Ensure {name} is a Pyomo Param, Var, Expression, or ScalarExpression."
                 )
+
+    def economy_of_numbers(
+        blk, cum_num_units, cost_FOAK, CE_index_year, learning_rate=0.04
+    ):
+        """
+        Equations for Economy of Numbers (EoN) derived from NETL Quality Guidelines for Energy System Studies
```
Agfritz, line 3451: It looks like this QGESS is no longer publicly released (but i do see the references to it in other documents). Were you able to find a version of it?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -3442,3 +3443,65 @@ def assert_Pyomo_object(b, name):
...
+        b = - log(1-R)/log(2)
+        where Y is the cost of the Nth-of-A-Kind (NOAK) of the equipment, A is the cost of the FOAK,
+        X is the cumulative number of units, b is the learning rate exponent, and R is the learning
+        rate constant.
+
+        Args:
+            cum_num_units: The cumulative number of units.
+            cost_FOAK: The cost of manufacturing the First-of-A-Kind equipment.
+            CE_index_year: year for cost basis, e.g., "2021" to use 2021 dollars
+            learning_rate: ranges between 0.01 - 0.06, depending on the level of maturity
```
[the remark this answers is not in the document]
Agfritz: In a review of the QGESS, it shows a range of 0.01-0.1. https://www.cmu.edu/epp/iecm/rubin/PDF%20files/2016/An%20Assessment%20of%20NETL%20Cost%20Estimation%20Methodology_20160429.pdf

would it make sense to increase the range and to provide a link to the table recommending learning rate by maturity level?

src/prommis/uky/costing/ree_plant_capcost.py
Akintomiwaojo: I just realized my version is missing from the official NETL page. However, the info is also in the review of the QGESS you shared earlier. Should I cite the review document or the QGESS report, considering that the review also cited the QGESS report?

bpaul4: See my comment above.

If you cite the QGESS, check that your version is publicly available or reference a newer version with the same information. Your version might be on OSTI or another database.

Akintomiwaojo: Unfortunately, it's also not on OSTI or any other database. Since it has the same information, I will use the CMU report. Thank you.

Agfritz: Ok thank you! It looks like the reference is fixed below but I would also recommend removing the reference to QGESS here since it is no longer published. I also did not see that the CMU report published the equations, were they in the report or did i miss them?

Akintomiwaojo: The CMU report also published the equations (Chapter 6, equations 2 and 3). However, the definition of variables in the report is slightly different from what I used in the code. For instance, they worked with a reduction in time as the cumulative production unit increased, and for ours, it's cost reduction.

Agfritz: The equation formation used is shown in this publication: https://www.osti.gov/pages/biblio/1868500
I would recommend adding this as the citation for the equations.

bpaul4 (approved):
I think this looks great! Thanks @Akintomiwaojo for contributing this.

Agfritz (approved):
Thanks for making all the changes so quickly Tomi, looks good to go!

PR #139, opened 2025-03-19
Adds two-salt diafiltration unit model

mollydougher:
## Summary/Motivation:
This PR adds a unit model and sample flowsheet for diafiltration system with two salts. This multi-component model for diafiltration takes lithium and cobalt as a case study, but there will be generalizations to additional solutes in future PRs.

- Adds a high-fidelity, DAE unit model for a multi-salt diafiltration separation.
- Provides an example flowsheet using this unit model.

### Limitations of the Model:
- Describes the transport of a two-cation system with a common anion, specifically lithium chloride and cobalt chloride.
- Omits the formation of a boundary layer due to concentration polarization, partitioning at the surface of the membrane, and assumes the membrane is uncharged.
- Approximates the ion-ion interactions using linearized expressions for cross-diffusion coefficients.

### Planned Improvements (left for future PRs, not necessarily in this order):
- Generalize the model to salt solutions other than lithium chloride and cobalt chloride.
- Extend the model to salt solutions with 2 cations and 2 anions.
- Extend the model to salt solutions with _m_ cations and _n_ anions.
- Include Donnan exclusion (i.e., incorporate the membrane partition coefficient, _H_, in the model)
- Include the boundary layer formation due to concentration polarization.

- [x] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,870 @@
...
+        self.retentate_conc_mass_chlorine = Var(
+            self.x_bar,
+            initialize=22.5,
+            units=units.kg / units.m**3,
+            domain=NonNegativeReals,
+            doc="Mass concentration of chlorine in the retentate, x-dependent",
+        )
+        self.permeate_flow_volume = Var(
+            self.x_bar,
+            initialize=0,
```
bpaul4, line 241: Is this expected to be a very small value? It's usually best to initialize at a non-zero value to maintain numerical stability while solving - the diagnostics in particular sometimes have an issue if variables take a zero value and end up in a denominator or exponential expression.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,870 @@
...
+
+        # other physical constraints
+        def _osmotic_pressure_calculation(self, x):
+            return self.osmotic_pressure[x] == units.convert(
+                (
+                    (
+                        self.config.property_package.num_solutes
+                        * Constants.gas_constant  # J / mol / K
+                        * 298
+                        * units.K  # assume room temp
```
bpaul4, line 623: Should this use the actual temperature state property?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,870 @@
...
+                0 * units.kg / units.m**3
+            )
+
+        self.initial_d_retentate_conc_mass_lithium_dx = Constraint(
+            rule=_initial_d_retentate_conc_mass_lithium_dx
+        )
+
+        def _initial_d_retentate_conc_mass_cobalt_dx(self):
+            return self.d_retentate_conc_mass_cobalt_dx[0] == (
+                0 * units.kg / units.m**3
```
bpaul4, line 847: As mentioned above, it's helpful to initialize with a very small non-zero value instead of zero.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@pytest.mark.unit
+def test_config(diafiltration_two_salt):
+    assert len(diafiltration_two_salt.fs.unit.config) == 17
+
+    assert not diafiltration_two_salt.fs.unit.config.dynamic
+    assert not diafiltration_two_salt.fs.unit.config.has_holdup
+
+    # TODO: add additional relevant assertions
```
bpaul4, line 124: Check that the config parameters have the correct values - see my earlier comment on building these inside the unit model instead of as config arguments, if you did that you would then add assertions in `test_build` below that each config argument creates a `Param` with the expected value.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class TestDiafiltrationTwoSalt(object):
+    @pytest.mark.build
+    @pytest.mark.unit
+    def test_build(self, diafiltration_two_salt):
+        assert hasattr(diafiltration_two_salt.fs.unit, "volume_flux_water")
+        # TODO: add additional relevant assertions
```
bpaul4, line 132: Check that all generated attributes exist - you can use `isinstance(x, Var)` to check that something exists and is the correct type.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @pytest.mark.component
+    def test_diagnostics(self, diafiltration_two_salt):
+        assert_units_consistent(diafiltration_two_salt.fs.unit)
```
bpaul4, line 136: This is redundant, the diagnostics will check for this too.

src/prommis/nanofiltration/diafiltration_solute_properties.py
```diff
@@ -0,0 +1,147 @@
...
+            self.component_list,
+            units=units.kg / units.mol,
+            initialize={
+                "Li": 0.006941,
+                "Co": 0.05893,
+                "Cl": 0.03545,
+            },
+        )
+
+        # add sigma
```
MarcusHolly, line 69: Maybe change this to "add osmotic coefficient, where 1 represents ideal behavior"?

src/prommis/nanofiltration/diafiltration_flowsheet_two_salt.py
```diff
@@ -0,0 +1,338 @@
...
+    dt.assert_no_structural_warnings()
+
+    solve_model(m)  # TODO: debug numerical scaling for higher NFE
+    dt.report_numerical_issues()
+
+    plot_results(m)
+    plot_membrane_results(m)
+
+
+def build_membrane_parameters(m):
```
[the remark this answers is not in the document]
MarcusHolly: Same question - seems like these should be mutable parameters inside the unit model

src/prommis/nanofiltration/diafiltration_two_salt.py
mollydougher: Yes, but only at the beginning of the membrane. At the very start of the membrane (x=0), the retentate flow is equal to the the feed inlet, thus there is no flux through the membrane and no permeate flow concentration. I can initialize these permeate variables at numerically 0, and likewise set the LBs to 1e-4.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
bpaul4: For the test, I think checking the value is useful, especially if the set value is different than the default value.

mollydougher:
@bpaul4 @MarcusHolly Besides the unit model documentation (which I am still working on), this is ready for another round of reviews.

src/prommis/nanofiltration/README.md
```diff
@@ -35,3 +17,27 @@ This file uses the nf_brine.py flowsheet to perform a sensitivity analysis on th
...
 
 4. Feed Pressure (bar) versus Volume Recovery: The optimal feed pressure increases as the system capacity increases and the membrane area reaches its upper bound.
+
+## diafiltration_properties.py
+This contains the property package for the sieving coefficient model used for separating lithium and cobalt ions via diafiltration.
+
+## diafiltration.py
+A fixed configuration nanofiltration membrane system with an additional, dilute inlet stream (the diafiltrate) is used to simulate a diafiltration cascade for recovering lithium and cobalt from spent lithium-ion battery leachate. The membrane model originates from Wamble et al. (https://pubs.acs.org/doi/full/10.1021/acssuschemeng.2c02862) and was initially implemented in Pyomo by Andrew Lee.
+
+A custom costing class is added to the flowsheet to minimize the annualized operating cost, subject to lithium and cobalt recovery bounds, the transport model, and mass balances. There are 3 DOF (the membrane length in each of the 3 stages).
```
MarcusHolly, line 27:
```suggestion
A custom costing class is added to the flowsheet to minimize the annualized operating cost, subject to lithium and cobalt recovery bounds, the transport model, and mass balances. There are 3 degrees of freedom (the membrane length in each of the 3 stages).
```
I would opt to spell this out as to not confuse newer users

src/prommis/nanofiltration/README.md
```diff
@@ -35,3 +17,27 @@ This file uses the nf_brine.py flowsheet to perform a sensitivity analysis on th
...
+
+## diafiltration_two_salt.py
+Multi-component (two-salt, common anion) unit model for diafiltration. The membrane is designed for use in the diafiltration cascade, i.e., is a spiral-wound membrane module. See source code for detailed documentation.
```
MarcusHolly, line 30: Embed a hyperlink to the source code inside of "source code"?

src/prommis/nanofiltration/diafiltration_flowsheet_two_salt.py
```diff
@@ -0,0 +1,242 @@
...
+    m.fs.membrane = TwoSaltDiafiltration(
+        property_package=m.fs.properties,
+        NFEx=5,
+        NFEz=5,
+    )
+
+    dt = DiagnosticsToolbox(m)
+    dt.assert_no_structural_warnings()
+
+    solve_model(m)  # TODO: debug numerical scaling for higher NFE
```
MarcusHolly, line 51: Has this TODO been done?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1028 @@
+#####################################################################################################
+# “PrOMMiS” was produced under the DOE Process Optimization and Modeling for Minerals Sustainability
+# (“PrOMMiS”) initiative, and is copyright (c) 2023-2025 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory, et al. All rights reserved.
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license information.
+#####################################################################################################
+r"""
+Two Salt Diafiltration Unit Model
```
MarcusHolly, line 8:
```suggestion
Two-Salt Diafiltration Unit Model
```

src/prommis/nanofiltration/README.md
```diff
@@ -1,31 +1,13 @@
...
-This contains the property package for the sieving coefficient model used for separating lithium and cobalt ions via diafiltration.
-
-### diafiltration.py
-A fixed configuration nanofiltration membrane system with an additional, dilute inlet stream (the diafiltrate) is used to simulate a diafiltration cascade for recovering lithium and cobalt from spent lithium-ion battery leachate. The membrane model originates from Wamble et al. (https://pubs.acs.org/doi/full/10.1021/acssuschemeng.2c02862) and was initially implemented in Pyomo by Andrew Lee.
-
-A custom costing class is added to the flowsheet to minimize the annualized operating cost, subject to lithium and cobalt recovery bounds, the transport model, and mass balances. There are 3 DOF (the membrane length in each of the 3 stages).
-
-### nf_brine.py
+## nf_brine.py
 This is an example flowsheet for the separation of lithium and magnesium ions using nanofiltration (NF). The unit model is the DSPM-DE and the property model is the MCAS, both of which can be found in WaterTAP. Currently, the inlet feed is modeling the Salar de Atacama brine.
```
bpaul4, line 4: It may be helpful to define these acronyms.

MarcusHolly (approved):
A couple of minor comments, but I'll approve regardless

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
```diff
@@ -0,0 +1,359 @@
...
+            ],
+        }
+
+        for model_result, test_val in test_dict_small_values.values():
+            assert pytest.approx(test_val, rel=1e-2) == value(model_result)
+
+    @pytest.mark.component
+    def test_numerical_issues(self, diafiltration_two_salt):
+        dt = DiagnosticsToolbox(diafiltration_two_salt.fs.unit)
+        dt.report_numerical_issues()  # there may be some variables at/near bound of zero (expected)
```
bpaul4, line 359: This is reporting the issues, not checking it. There are flags to ignore potential evaluation errors or inconsistent units, however it would be better to address the near-bound values if possible.

bpaul4 (commented):
This is looking very nice, just a few comments and I agree with Marcus's suggestions as well. Please request my review when the constraints part of the docstring is ready.

src/prommis/nanofiltration/README.md
```diff
@@ -35,3 +17,27 @@ This file uses the nf_brine.py flowsheet to perform a sensitivity analysis on th
 3. Membrane Area (m2) versus Volume Recovery: The membrane area increases as the system capacity (volume recovery) increases.
```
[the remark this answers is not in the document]
mollydougher: This doesn't seem to be working; I will revisit this in a subsequent PR.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
mollydougher: @bpaul4 The current issue I am having with this line of code is when I include `assert_no_numerical_warnings()` the test runs without errors on my local machine(s), but the GitHub pytest checks failed. I think its a scaling issue but I am unsure how to resolve.

src/prommis/nanofiltration/diafiltration_flowsheet_two_salt.py
mollydougher: @MarcusHolly This is still a work in progress. My plan was to investigate finer discretization points, and possibly alternative discretization methods, in a subsequent PR.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
bpaul4: When I encounter this type of behavior, it's usually a package difference in the Python environment. Could you please check your local environment against what was installed in https://github.com/prommis/prommis/actions/runs/14110273522/job/39527019801 and see if any significant packages are different, especially `numpy`?

mollydougher: @bpaul4 I am using an old version of `numpy`. The job is using `2.0.2` while I have `1.21.5` installed in my environment. I am also using an older version of `pyomo` (`6.4.4` versus `6.9.1`). I'll update these.

mollydougher: @bpaul4 I believe all the significant packages have the same version numbers (`numpy`,`scipy`,`idaes`,`pytest`), but there is still a failure in the job that does not get triggered in my local environment.

bpaul4: Could you please check `watertap` as well? Perhaps there is a difference in the solver that is causing the issue.

If WaterTAP isn't the issue, I will download it and try running it locally to see if it passes for me. @aostace01 do you have any additional suggestions to look into?

mollydougher: Both are using `watertap-1.3.0` and `watertap-solvers-24.12.9`

bpaul4: It appears that the mass flux of cobalt is rounding down to zero for some points, and this is causing the diagnostics issue:

```
----------------------------------------------- Captured stdout call -----------------------------------------------
====================================================================================
Model Statistics

    Jacobian Condition Number: 1.555E+13

------------------------------------------------------------------------------------
1 WARNINGS

    WARNING: 4 Variables at or outside bounds (tol=0.0E+00)

------------------------------------------------------------------------------------
6 Cautions

    Caution: 81 Variables with value close to their bounds (abs=1.0E-04, rel=1.0E-04)
    Caution: 89 Variables with value close to zero (tol=1.0E-08)
    Caution: 216 Variables with extreme value (<1.0E-04 or >1.0E+04)
    Caution: 41 Constraints with mismatched terms
    Caution: 24 Constraints with potential cancellation of terms
    Caution: 547 extreme Jacobian Entries (<1.0E-04 or >1.0E+04)

------------------------------------------------------------------------------------
Suggested next steps:

    display_variables_at_or_outside_bounds()

====================================================================================
====================================================================================
The following variable(s) have values at or outside their bounds (tol=0.0E+00):

    fs.unit.mass_flux_cobalt[0.2] (free): value=0.0 bounds=(0, None)
    fs.unit.mass_flux_cobalt[0.4] (free): value=0.0 bounds=(0, None)
    fs.unit.mass_flux_cobalt[0.6] (free): value=0.0 bounds=(0, None)
    fs.unit.mass_flux_cobalt[0.8] (free): value=0.0 bounds=(0, None)

====================================================================================
mass_flux_cobalt : Mass flux of cobalt across the membrane (z-direction, x-dependent)
    Size=6, Index=fs.unit.x_bar, Units=kg/h/m**2
    Key : Lower : Value                  : Upper : Fixed : Stale : Domain
      0 :     0 : 1.8716366192010145e-10 :  None : False : False : NonNegativeReals
    0.2 :     0 :                    0.0 :  None : False : False : NonNegativeReals
    0.4 :     0 :                    0.0 :  None : False : False : NonNegativeReals
    0.6 :     0 :                    0.0 :  None : False : False : NonNegativeReals
    0.8 :     0 :                    0.0 :  None : False : False : NonNegativeReals
      1 :     0 :  5.396300503818238e-11 :  None : False : False : NonNegativeReals
mass_flux_lithium : Mass flux of lithium across the membrane (z-direction, x-dependent)
    Size=6, Index=fs.unit.x_bar, Units=kg/h/m**2
    Key : Lower : Value                 : Upper : Fixed : Stale : Domain
      0 :     0 : 7.796842866972918e-10 :  None : False : False : NonNegativeReals
    0.2 :     0 : 6.529287396850113e-08 :  None : False : False : NonNegativeReals
    0.4 :     0 : 6.535290567311862e-08 :  None : False : False : NonNegativeReals
    0.6 :     0 : 6.536896769146717e-08 :  None : False : False : NonNegativeReals
    0.8 :     0 : 6.537404748016172e-08 :  None : False : False : NonNegativeReals
      1 :     0 : 6.537532957890781e-08 :  None : False : False : NonNegativeReals
```

mollydougher: Interesting. My solution looks like this for `mass_flux_cobalt`:
```
mass_flux_cobalt : Mass flux of cobalt across the membrane (z-direction, x-dependent)
    Size=6, Index=fs.membrane.x_bar, Units=kg/h/m**2
    Key : Lower : Value                  : Upper : Fixed : Stale : Domain
      0 :     0 : 2.1822480705943705e-10 :  None : False : False : NonNegativeReals
    0.2 :     0 : 5.1250137768213775e-11 :  None : False : False : NonNegativeReals
    0.4 :     0 : 5.0751563934850483e-11 :  None : False : False : NonNegativeReals
    0.6 :     0 :  5.107298342321195e-11 :  None : False : False : NonNegativeReals
    0.8 :     0 :  5.287117508434255e-11 :  None : False : False : NonNegativeReals
      1 :     0 :   6.93549579179516e-11 :  None : False : False : NonNegativeReals
```

bpaul4: @dallan-keylogic do you have a suggestion on this, possibly related to addressing @aostace01's point on restoration below?

mollydougher:
@bpaul4 Re-requesting your review now that the constraint documentation is complete.

bpaul4:
Thanks @mollydougher, could you please update https://github.com/prommis/prommis/blob/main/docs/api.rst? This will include your new model in the API documentation generated by the CI run.

bpaul4:
@mollydougher the autodoc summary looks great, thanks for writing it. Everything else looks great, too.

What is the plan to address the numerical diagnostics issue?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1097 @@
...
+.. math:: q_r(\bar{x}=0)=q_f+q_d
+.. math:: q_r(\bar{x}=0) c_{\mathrm{Li^+},r}(\bar{x}=0)=q_f c_{\mathrm{Li^+},f} + q_d c_{\mathrm{Li^+},d}
+.. math:: q_r(\bar{x}=0) c_{\mathrm{Co^{2+}},r}(\bar{x}=0)=q_f c_{\mathrm{Co^{2+}},f} + q_d c_{\mathrm{Co^{2+}},d}
+.. math:: q_p(\bar{x}=0) = 0
+.. math:: c_{\mathrm{Li^+},p} (\bar{x}=0) = 0
+.. math:: c_{\mathrm{Co^{2+}},p} (\bar{x}=0) = 0
+.. math:: \frac{\mathrm{d}q_r(\bar{x})}{\mathrm{d}\bar{x}}(\bar{x}=0)=0
+.. math:: \frac{\mathrm{d}c_{\mathrm{Li^+},r}(\bar{x})}{\mathrm{d}\bar{x}}(\bar{x}=0)=0
+.. math:: \frac{\mathrm{d}c_{\mathrm{Co^{2+}},r}(\bar{x})}{\mathrm{d}\bar{x}}(\bar{x}=0)=0
+
```
dallan-keylogic, line 208: @lbianchi-lbl , is it possible to move this documentation to a `.rst` file? I'm personally not a fan of extensive documentation inside a file or function docstring, but I don't know how PrOMMiS's documentation is set up, nor which Diataxis category to put this in.

(This is minor feedback, and my mind is open to being changed about documentation organization.)

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1097 @@
...
+            mutable=True,
+            units=units.m,
+            doc="Length of the membrane, wound radially",
+        )
+        self.applied_pressure = Param(
+            initialize=10,
+            mutable=True,
+            units=units.bar,
+            doc="Pressure applied to membrane",
+        )
```
dallan-keylogic, line 316: Are these parameters going to be optimized? If so, they should be `Vars`.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1097 @@
...
+            mutable=True,
+            units=units.kg / units.m**3,
+            doc="Mass concentration of lithium in the feed",
+        )
+        self.feed_conc_mass_cobalt = Param(
+            initialize=17,
+            mutable=True,
+            units=units.kg / units.m**3,
+            doc="Mass concentration of cobalt in the feed",
+        )
```
dallan-keylogic, line 340: These also need to be `Vars`, and they need aliases on a port (using `References`) that are compatible with the provided property package.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1097 @@
...
+            mutable=True,
+            units=units.kg / units.m**3,
+            doc="Mass concentration of lithium in the diafiltrate",
+        )
+        self.diafiltrate_conc_mass_cobalt = Param(
+            initialize=0.2,
+            mutable=True,
+            units=units.kg / units.m**3,
+            doc="Mass concentration of cobalt in the diafiltrate",
+        )
```
dallan-keylogic, line 358: I am confused about which stream is the "diafiltrate". I thought it was the permeate, but it doesn't make sense to specify the permeate concentration as a `Param`. The term "diafiltrate" also doesn't appear in the membrane schematic. Is there an inlet flow on the permeate side in this model, or just an outlet flow?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1097 @@
...
+        self.temperature = Param(
+            initialize=298,
+            mutable=True,
+            units=units.K,
+            doc="System temperature",
+        )
```
dallan-keylogic, line 364: Is this going to be optimized? Then it should be a `Var`. In the long run, it might be good to put it on the property package, but, since no energy balance exists, we should avoid that at present.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1097 @@
...
+
+    def discretize_model(self):
+        discretizer = TransformationFactory("dae.finite_difference")
+        discretizer.apply_to(
+            self, wrt=self.x_bar, nfe=self.config.NFEx, scheme="FORWARD"
+        )
+        discretizer.apply_to(
+            self, wrt=self.z_bar, nfe=self.config.NFEz, scheme="FORWARD"
+        )
```
dallan-keylogic, line 1063: We're flowing from x=0 to x=1 and z=0 to z=1, right? Why are we using a forward difference instead of a backward difference?

dallan-keylogic (changes requested):
At present, this model has no ports, so it cannot be used with other unit models. Ports need to be added, and the flowsheet model should be modified to specify the inlet stream in a `Feed` unit model, connect it with an arc to the diafiltration unit, then connect the diafiltration unit to a `Product` unit model with another arc. That way we can test that the property package works (since it isn't used to create any `StateBlocks` in the diafiltration unit model.

src/prommis/nanofiltration/diafiltration_two_salt.py
mollydougher: The diafiltrate is an additional (optional) feed stream that can be concentrated or dilute. It helps to mitigate concentration effects at the surface of the membrane. Right now the diafiltrate stream is lumped with the feed and used as an initial condition for the retentate. When I add Ports to the model, I will separate the diafiltrate and feed streams.

I also updated the membrane schematic to reflect this clarification.

dallan-keylogic: Thank you for clarifying. In the future, it might be better to combine the two streams in a `Mixer` block before they reach the diafiltration block, but at present having separate ports for both inputs will work.

src/prommis/nanofiltration/diafiltration_two_salt.py
mollydougher: Yes, thank you for catching this. I called them `Param`s here, but really they are `Var`s I fix for simulation. Is it preferred to fix these `Var`s within the unit model, such that the unit model has 0 DOF and a user unfixes them for optimization, or leave them unfixed within the unit model (non-zero DOF) and require the user to set these `Var`s in a simulation?

dallan-keylogic: Generally, you should leave them unfixed in the unit model and force the user to fix them for simulation. In this case, it wouldn't matter too much, but you run into problems if the `Var` you're trying to fix is indexed by a `ContinuousSet` (usually time) that is then discretized with Pyomo DAE. When new `Vars` are created at the new time indices created by discretizing the continuous set, Pyomo doesn't know to fix these new `Vars` or even to propagate the variable value at other time indices.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -259,6 +264,27 @@ class TwoSaltDiafiltrationData(UnitModelBlockData):
...
+            doc="Length of the membrane, wound radially in meters",
+        ),
+    )
+    CONFIG.declare(
+        "applied_pressure",
+        ConfigValue(
+            default=10,
+            doc="Pressure applied to membrane in bar",
+        ),
+    )
```
dallan-keylogic, line 289: These don't need to be config entries. You can just initialize the `Vars` at those values.

mollydougher:
> @mollydougher the autodoc summary looks great, thanks for writing it. Everything else looks great, too.
> 
> What is the plan to address the numerical diagnostics issue?

@bpaul4 The combination of using backward-backward discretization, updating the tolerances for values I expect to be zero (making them numerically zero), and setting specific constraints for the fluxes I expect to be zero (at x=0) seems to have resolved the numerical diagnostics issues.

dallan-keylogic (approved):
LGTM

dallan-keylogic:
Thank you @aostace01 for catching that. However, I disagree about making `nfe` a flowsheet-level variable.

Typically how we deal with this situation is to not have the property state variables indexed by space and time, but rather create a `StateBlock` that is indexed by state and time. So instead of accessing `properties.flow_mol[t, x]` we access `properties[t, x].flow_mol`. See the `ControlVolume1D` [as an example.](https://github.com/IDAES/idaes-pse/blob/6a588c799fbaf00cb1b32b814fcee6ad111bc4c1/idaes/core/base/control_volume1d.py#L348) Since that's a pretty complex unit model, let me know if you have any questions.

src/prommis/nanofiltration/diafiltration_flowsheet_two_salt.py
```diff
@@ -0,0 +1,394 @@
...
+    # add feed blocks for feed and diafiltrate
+    m.fs.feed_block = Feed(property_package=m.fs.feed_properties)
+    m.fs.diafiltrate_block = Feed(property_package=m.fs.feed_properties)
+
+    # add the membrane unit model
+    m.fs.membrane = TwoSaltDiafiltration(
+        property_package=m.fs.properties,
+        NFEx=10,
+        NFEz=5,
+    )
```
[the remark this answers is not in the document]
mollydougher: Thank you for your comment @aostace01. I will start with @dallan-keylogic's suggestion to address this.

With regard to nfe=10 being lucky, I believe this is due to the two independent discretizations you mentioned, which use slightly different methods, so there is a significant digit issue between the ports.

mollydougher:
@aostace01 @dallan-keylogic I believe the aforementioned bug has been addressed. Please let me know if you have any remaining comments.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -0,0 +1,1116 @@
...
+        # add components
+        self.solutes = Set(initialize=["Li", "Co", "Cl"])
+
+        # add global variables
+        self.membrane_width = Var(
+            initialize=1,
+            units=units.m,
+            domain=NonNegativeReals,
+            doc="Width of the membrane (x-direction)",
+        )
```
dallan-keylogic, line 372: Generally we prefer to set `bounds=[0, None]` instead of a domain. There are situations where we'd like to get rid of the lower bound of a `Var` in order to avoid issues with IPOPT, but `x.setlb(None)` isn't sufficient. First you need to override the domain, _then_ modify the lower bound. `x.setlb(None)` silently fails, and you'll find `x.bounds == [0, None]` if you don't correct the domain first.

dallan-keylogic (changes requested):
Something I hadn't noticed until this recent changeset is that you're not putting individual streams for the retentate and permeate, but the entire flow and concentration profile across the entire length of the membrane. Is the plan to stack these membrane models lengthwise? If so, the inlet port should also be indexed by length, and you should then create terminal channel models. 

If you're not stacking them lengthwise, you should add a permeate material balance, sum up all the material flows on the permeate side, then add a port with an outlet stream that is not indexed by length.

mollydougher:
> Something I hadn't noticed until this recent changeset is that you're not putting individual streams for the retentate and permeate, but the entire flow and concentration profile across the entire length of the membrane. Is the plan to stack these membrane models lengthwise? If so, the inlet port should also be indexed by length, and you should then create terminal channel models.
> 
> If you're not stacking them lengthwise, you should add a permeate material balance, sum up all the material flows on the permeate side, then add a port with an outlet stream that is not indexed by length.

@dallan-keylogic Thank you for catching this. There may be some discussion further down the line, but currently the intention is to stage this membrane model such that the permeate of the previous membrane is the feed/retentate inlet of the next. The permeate mass balances already existed, but I have updated the Port References to reflect that the permeate and retentate product streams are collected at the end of the membrane module. Please let me know if there are any remaining concerns.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -1096,27 +1096,65 @@ def add_scaling_factors(self):
...
+        self.retentate_outlet.add(
+            self._retentate_conc_mass_cobalt_ref, "conc_mass_cobalt"
+        )
+        self._retentate_conc_mass_chlorine_ref = Reference(
+            self.retentate_conc_mass_comp[0, value(self.membrane_width), "Cl"]
+        )
+        self.retentate_outlet.add(
+            self._retentate_conc_mass_chlorine_ref, "conc_mass_chlorine"
+        )
```
dallan-keylogic, line 1136: You didn't need to break `conc_mass_comp` into element-by-element variables. You can create `References` to slices. For example, you could create `self._retentate_conc_mass_comp_ref = Reference(self.retentate_conc_mass_comp[:, self.x_bar.last(), :])` to collapse the length dimension while retaining the time and component dimensions.

You should make sure to use `x_bar.last()` , because that's always going to give the right answer. You're working with a reduced length domain, and `membrane_width` is a `Var`, so `value` may not even return a member of `x_bar`. (I think that's a moot point in this case because `membrane_width` is initialized to 1 and probably can't be changed by the user before these ports are created, but it's still bad form).

dallan-keylogic (approved):
Alright, looks good. Thank you for making changes.

ksbeattie:
@aostace01, can you take a look that this PR?

PR #143, opened 2025-04-04
Location factor

Lingyan90:
## Addresses Issue: 
Added location factor to account for CAPEX at different market. 

## Summary/Motivation:
This update addresses Andrew Lee’s earlier comment regarding the inclusion of different currencies. Instead of incorporating multiple currencies, we apply a location factor, which offers greater stability over time. Additionally, this approach aligns with the scope of our analysis, which focuses on a single market rather than a global one.

- Added `location_factor.json` file. This file stored 139+ locations worldwide.
- Loaded the `location_factor.json` file into the `costing_dictionaries.py`.
- Imported location factor in `ree_plant_capcost.py` and updated the `total_BEC` with location factor considered.
- Tested the location factor adjusted total BEC in `test_ree_costing.py`. Four possible cases are been tested.

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/uky/costing/location_factors.json
```diff
@@ -0,0 +1,1694 @@
...
+        "city": null,
+        "location_factor": {
+            "min": 0.9,
+            "max": 0.94,
+            "average": 0.92
+        }
+    },
+    {
+        "country": "Australia",
+        "city": "Melborne",
```
bpaul4, line 58:
```suggestion
        "city": "Melbourne",
```

src/prommis/uky/costing/location_factors.json
```diff
@@ -0,0 +1,1694 @@
...
+        "city": null,
+        "location_factor": {
+            "min": 0.9,
+            "max": 0.95,
+            "average": 0.925
+        }
+    },
+    {
+        "country": "Canada",
+        "city": "Cagary AL",
```
bpaul4, line 247:
```suggestion
        "city": "Calgary AL",
```

src/prommis/uky/costing/location_factors.json
```diff
@@ -0,0 +1,1694 @@
...
+        "city": "Fort McMurray AL",
+        "location_factor": {
+            "min": 1.4,
+            "max": 1.55,
+            "average": 1.475
+        }
+    },
+    {
+        "country": "Canada",
+        "city": "Hatifax NS",
```
bpaul4, line 283:
```suggestion
        "city": "Halifax NS",
```

src/prommis/uky/costing/location_factors.json
```diff
@@ -0,0 +1,1694 @@
...
+        "city": "St Johns NFL",
+        "location_factor": {
+            "min": 0.96,
+            "max": 0.98,
+            "average": 0.97
+        }
+    },
+    {
+        "country": "Canada Canada",
+        "city": "Toronto ONT Whitehorse YUK",
```
bpaul4, line 382:
```suggestion
        "country": "Canada",
        "city": "Toronto ONT",
```

src/prommis/uky/costing/location_factors.json
```diff
@@ -0,0 +1,1694 @@
...
+        "country": "Costa Rica",
+        "city": null,
+        "location_factor": {
+            "min": 0.92,
+            "max": 0.94,
+            "average": 0.93
+        }
+    },
+    {
+        "country": "Cote D'lvory",
```
bpaul4, line 534:
```suggestion
        "country": "Cote D' Ivory",
```

src/prommis/uky/costing/location_factors.json
```diff
@@ -0,0 +1,1694 @@
...
+        "country": "Nepal",
+        "city": null,
+        "location_factor": {
+            "min": 0.89,
+            "max": 0.94,
+            "average": 0.915
+        }
+    },
+    {
+        "country": "Netherlands",
```
bpaul4, line 1182: To be consistent with the table.

```suggestion
        "country": "The Netherlands",
```

src/prommis/uky/costing/location_factors.json
```diff
@@ -0,0 +1,1694 @@
...
+        "country": "Oman",
+        "city": null,
+        "location_factor": {
+            "min": 0.93,
+            "max": 0.98,
+            "average": 0.955
+        }
+    },
+    {
+        "country": "Panarna",
```
bpaul4, line 1263:
```suggestion
        "country": "Panama",
```

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2710,6 +2716,11 @@ def get_total_BEC(b, CE_index_year, watertap_blocks=None):
                 f"Valid CE index options include CE500, CE394 and years from "
                 f"1990 to 2020."
             )
+        # Define the lists to ensure hit no errors during testing
+        b.BEC_list = []
+        b.custom_fixed_costs_list = []
+        b.custom_variable_costs_list = []
+        b.watertap_fixed_costs_list = []
```
bpaul4, line 2723: These are already defined on lines 494 - 499, I don't think we need to define them again here. Can we remove these?

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2777,9 +2788,63 @@ def get_total_BEC(b, CE_index_year, watertap_blocks=None):
...
+                }
+            )
+
+            available_city_list = (
+                ", ".join(available_cities) if available_cities else "None"
+            )
+            raise AttributeError(
+                f"No location factor found for {location}. "
+                f"Available cities for '{location[0]}' are: {available_city_list}. "
+                f"Did you mean one of those?"
```
bpaul4, line 2828:
```suggestion
                f"Did you mean one of these?"
```

src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -2777,9 +2788,63 @@ def get_total_BEC(b, CE_index_year, watertap_blocks=None):
...
+            )
+
+        # No country match
+        else:
+            available_countries = sorted(
+                {loc[0] for loc in location_factor_dict.keys()}
+            )
+            raise AttributeError(
+                f"No location factor found for country '{location[0]}'. "
+                f"Available countries are: {', '.join(available_countries)}"
```
bpaul4, line 2838: This is a long list to print to the console. Should we point users to the JSON file instead?

I think for cities it's fine to list it, since each country only has between zero and five or so cities available.

src/prommis/uky/costing/tests/test_ree_costing.py
```diff
@@ -1399,6 +1400,87 @@ def test_costing_bounding_rerun_recalculate(self, model):
                 expected_costing_upper_bound[key], rel=1e-4
             )
 
+    @pytest.mark.unit
+    def test_location_factor(self):
```
bpaul4, line 1404: A couple comments

1) Have separate tests for test_location_factor (Test 1, defaults and check that solve is optimal), test_location_factor_invalidcountry (Test 2), test_location_factor_nocityprovided (Test 3), test_location_factor_invalidcity (Test 4)

2) While calling `get_total_BEC` is correct in terms of testing the functionality, it would be better to call `build_process_costs` instead since that is what users would do on their flowsheets. I think `test_REE_costing_useLangfactor` is probably a good template for you tests as it is the simplest case of calculating the total BEC using CAPEX from a unit model block, the argument `location` would need to be added to the `build_process_costs` call right before the `Lang_factor` is set.

3) Update the costing module:

In the method arguments starting on line 284 for `build_process_costs`, add `location=("United States", "Washington DC")` right after `total_purchase_cost`

In the docstring arguments list starting on line 364, add `location` and a short description

line 502 update to `self.get_total_BEC(CE_index_year, watertap_blocks, location_factor=location_factor)

bpaul4 (commented):
Thanks @Lingyan90, the heart of the new code looks fantastic and the dictionary is very thorough. I have a few suggestions for the dictionary and for the tests.

src/prommis/uky/costing/location_factors.json
Lingyan90: nice catches, doubled checked all country, city, and province names, corrected all that's incorrect

src/prommis/uky/costing/ree_plant_capcost.py
Lingyan90: since the list was above, it makes more sense to use `those` than `these` I think. If use `these`, the list should be below a sentence like `Did you mean one of these:....?`

MarcusHolly (commented):
The json file itself looks great. I agree with Brandon's comment regarding the testing, so I'll re-review once his comments are addressed

src/prommis/uky/costing/ree_plant_capcost.py
Lingyan90: Yes, If use the way you suggested to build the model via `build_process_costs` other than calling `get_total_BEC` in test, it'll pass the test, if calling `get_total_BEC` directly, it'll need to define the list there.

I've modified the model build method to use `build_process_costs` and removed the above list definition.

bpaul4:
This update will be part of #150, and this PR may be closed in favor of one in IDAES. This will remain a draft while offline discussion continues.

src/prommis/uky/costing/costing_dictionaries.py
```diff
@@ -265,3 +266,37 @@ def load_default_resource_prices():
...
+def load_location_factor():
+    """
+    Estimate the cost of constructing the same plant in different global locations using location factors.
+
+    This method uses a location (or investment site) factor to adjust the total permanent investment (TPI)
+    based on regional differences in labor costs, workforce efficiency, local regulations and customs,
+    union status, and other local economic conditions.
+
+    Reference:
+    Seider, Warren D., et al. *Product and Process Design Principles: Synthesis, Analysis, and Evaluation.*
```
MAZamarripa, line 280: full reference

lbianchi-lbl:
This should be left as-is until the corresponding PR has been opened in IDAES/idaes-pse.

PR #148, opened 2025-04-28
Replace Bare Exceptions

bpaul4:
## Summary/Motivation:
Better way to checking for reference objects in the net present value method.

- Replace bare exceptions with alternative syntax to conform to codebase standards.

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.

### DIFF
--- a/src/prommis/uky/costing/ree_plant_capcost.py
+++ b/src/prommis/uky/costing/ree_plant_capcost.py
@@ -33,6 +33,7 @@ from sys import stdout
 
 from pyomo.common.config import ConfigValue, ListOf
 from pyomo.common.dependencies import attempt_import
+from pyomo.core.base.component import Component
 from pyomo.core.base.expression import ScalarExpression
 from pyomo.core.base.units_container import InconsistentUnitsError, UnitsError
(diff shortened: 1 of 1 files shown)


src/prommis/uky/costing/ree_plant_capcost.py
```diff
@@ -3660,11 +3660,10 @@ def verify_calculate_from_inputs(b):
 
             for key in costs.keys():
                 # check if the object is a Reference
-                try:
-                    costs[key] = costs[key][None]
-                except:
-                    # continue on
-                    pass
+                if type(costs[key]) not in [float, int]:
```
dallan-keylogic, line 3663: Checking types like this is rather brittle. For example, this method would fail if `costs[key]` was a numpy float of some sort. The `isinstance` function checks whether a variable is a member of a class or any of its subclasses and is thus more robust for things like this. I'd `from pyomo.core.base.component import Component` and then check `isinstance(costs[key], Component)` to see if the variable is a Pyomo modeling component.

dallan-keylogic (approved):
LGTM

MarcusHolly (approved):

PR #164, opened 2025-07-14
Adds multi-component partitioning to diafiltration

mollydougher:
## Summary/Motivation:
This PR extends the mathematical model in `diafiltration_two_salt.py` to incorporate the multi-component partitioning behavior at the solution-membrane interfaces. Additionally, this PR adds the functionality to account for a fixed surface charge on the membrane, which is typically seen in nanofiltration membranes such as NF270.

- Adds solute partition coefficients to the two-salt diafiltration model
- Adds a non-zero fixed membrane charge to the two-salt diafiltration model
- Adjusts the membrane solute flux equation(s) to account for the extensions to the mathematical model described above
- Converts the diafiltration model from a mass-basis to a mole-basis
- Implements analytical expressions for the derived convection and diffusion coefficients
- Updates documentation

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
```diff
@@ -486,47 +613,47 @@ def test_solve(self, diafiltration_two_salt):
         test_dict = {
             "retentate_final": [
                 value(diafiltration_two_salt.fs.unit.retentate_flow_volume[0, 1]),
-                100.00000000030005,
+                63.0068903198034,
```
bpaul4, line 616: Is this change expected?

src/prommis/nanofiltration/diafiltration_solute_properties.py
```diff
@@ -58,32 +57,40 @@ def build(self):
...
             units=units.dimensionless,
             initialize={
-                "Li": 1,
-                "Co": 1,
-                "Cl": 1,
+                "Li": 0.3,
+                "Co": 0.3,
+                "Cl": 0.3,
             },
+            domain=NonNegativeReals,
```
dallan-keylogic, line 82: I *strongly* discourage using Pyomo domains for anything other than specifying integer variables. Otherwise, Pyomo silently truncates the bounds to be within the domain if the user tries to change them. Even if we expect a variable to always be nonnegative, we might want to relax the lower bound in order to prevent IPOPT from treating the lower bound as an active constraint.

src/prommis/nanofiltration/diafiltration_stream_properties.py
```diff
@@ -98,18 +97,15 @@ def build(self):
...
+            units=units.mol / units.m**3,
             initialize=1e-5,
             bounds=(1e-20, None),
         )
 
     def get_material_flow_terms(self, p, j):
-        return self.flow_vol * self.conc_mass_comp[j]
-
-    def get_material_flow_basis(self):
-        return MaterialFlowBasis.mass
```
dallan-keylogic, line 112: Shouldn't you update this to `MaterialFlowBasis.mole`?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -10,12 +10,12 @@
 
 Author: Molly Dougher
 
-This multi-component model for the diafiltration membrane model is a two-salt system with a common anion. The membrane is designed for use in the diafiltration cascade, i.e., is a spiral-wound membrane module.
+This membrane unit model is for the multi-component diafiltration of a two-salt system with a common anion. The membrane is designed for use in a diafiltration cascade, i.e., represents a spiral-wound membrane module.
```
dallan-keylogic, line 13: > i.e., represents a spiral-wound membrane module

Does this mean this membrane unit model can be chained together lengthwise to represent different layers in a spiral-wound membrane, or just that they can be chained together end-to-end for multi-stage diafiltration?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -24,28 +24,30 @@
...
+#. the lithium concentration in the diafiltrate (``diafiltrate_conc_mol_comp[0,"Li"]``; :math:`14 \, \mathrm{mol} \, \mathrm{m}^{-3}`)
+#. the cobalt concentration in the diafiltrate (``diafiltrate_conc_mol_comp[0,"Co"]``; :math:`3 \, \mathrm{mol} \, \mathrm{m}^{-3}`)
 
 Model Structure
 ---------------
 
-There are three phases in the Two-Salt Diafiltration model: the retentate, the membrane, and the permeate. The retentate and the permeate are only discretized with respect to :math:`x`, while the membrane is discretized with respect to both :math:`x` and :math:`z`. The resulting system of partial differential algebraic equations is solved by discretizing with the forward finite element method.
+There are three phases in the Two-Salt Diafiltration model: the retentate, the membrane, and the permeate. The retentate and the permeate are only discretized with respect to :math:`x`, while the membrane is discretized with respect to both :math:`x` and :math:`z`. The resulting system of partial differential algebraic equations is solved by discretizing with the backward finite element method.
```
dallan-keylogic, line 43: > backward finite element method.

Pyomo DAE does not have any finite element methods implemented. Are you using the backward finite *difference* method?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -24,28 +24,30 @@
...
 
 Assumptions
 -----------
 
-The partition coefficients of all ions in solution are equal to 1, meaning the concentration of an ion :math:`i` just outside the membrane is equal to the concentration of ion :math:`i` just inside the membrane. We also assume that the membrane has no surface charge, which is a valid assumption for NF90 membranes.
+The partition coefficients, which describe how the solutes partition across the solution-membrane interfaces, are derived considering steric and electrostatic hindrance.
```
dallan-keylogic, line 48: Are these partition coefficients equilibrium relationships or rate-based relationships?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -60,12 +62,12 @@
...
 .. math:: \mathcal{I}=\{\mathrm{Li^+,Co^{2+},Cl^-}\}
 
-There are 2 continuous sets for each length dimension: the :math:`x`-direction parallel to the membrane surface (width) and the :math:`z`-direction perpendicular to the membrane surface (thickness). :math:`x` and :math:`z` are non-dimensionalized with the membrane width (:math:`w`) and thickness (:math:`l`), respectively, to improve numerics.
+There are 2 continuous sets for each length dimension: the :math:`x`-direction parallel to the membrane surface and the :math:`z`-direction perpendicular to the membrane surface. :math:`x` and :math:`z` are non-dimensionalized with the module length or membrane width (:math:`w`) and membrane thickness (:math:`l`), respectively, to improve numerics.
 
 .. math:: \bar{x} \in \mathbb{R} \| 0 \leq \bar{x} \leq 1
 .. math:: \bar{z} \in \mathbb{R} \| 0 \leq \bar{z} \leq 1
 
-Some variables are discretized over time to be compatible with the property package. Thus, the following set is defined for time.
+Some variables are discretized over time to be compatible with the property package, even though this is not a dynamic model. Thus, the following set is defined for time.
```
dallan-keylogic, line 70: > discretized over time

You probably mean "have a time domain".

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -364,212 +407,226 @@ def add_variables(self):
...
-            bounds=[0, None],
+            bounds=[1e-11, None],
             doc="Length of the membrane, wound radially",
         )
         self.applied_pressure = Var(
-            initialize=10,
+            initialize=15,
             units=units.bar,
-            bounds=[0, None],
+            bounds=[1e-11, None],
```
dallan-keylogic, line 424: @agarciadiego , do we have the IPOPT option `bound_push=0` for this repo's default settings?

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -692,16 +747,31 @@ def _lumped_water_flux(self, x):
         self.lumped_water_flux = Constraint(self.x_bar, rule=_lumped_water_flux)
 
         def _D_lithium_lithium_calculation(self, x, z):
-            params_lithium_lithium = [-3.79e-06, -3.64e-08, 7e-09]
+            if value(self.membrane_fixed_charge) == 0:
+                # these params assume chi=0 mM and lithium & cobalt mem_conc range 50-80 mM
+                self.diffusion_params_lithium_lithium = [-4.07e-06, -3.99e-09, 4.01e-09]
+            if value(self.membrane_fixed_charge) == -140:
```
dallan-keylogic, line 753: I always get nervous when using exact equality with floating point numbers. At the very least, you should have an `else` clause for the case when `self.membrane_fixed_charge != 0 and self.membrane_fixed_charge != -140`.

Also, you shouldn't be creating a block attribute inside a rule for generating a `Constraint` or `Expression`. Furthermore, you shouldn't be using `self` as a function argument because it obscures what's actually happening inside the function---you're creating a new variable "self" that shadows the `self` keyword from the outer scope. Much better to use `b` or `blk` when defining a rule.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -710,16 +780,31 @@ def _D_lithium_lithium_calculation(self, x, z):
         )
 
         def _D_lithium_cobalt_calculation(self, x, z):
-            params_lithium_cobalt = [-2.4e-07, -9.4e-08, 1.82e-08]
+            if value(self.membrane_fixed_charge) == 0:
+                # these params assume chi=0 mM and lithium & cobalt mem_conc range 50-80 mM
+                self.diffusion_params_lithium_cobalt = [-9.62e-07, -1.03e-08, 1.04e-08]
+            if value(self.membrane_fixed_charge) == -140:
+                # these params assume chi=-140 mM and lithium mem_conc range 50-80 mM & cobalt mem_conc range 80-110 mM
+                self.diffusion_params_lithium_cobalt = [-1.63e-06, -1.10e-08, 1.33e-08]
```
dallan-keylogic, line 788: Same comment as above.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -728,16 +813,31 @@ def _D_lithium_cobalt_calculation(self, x, z):
         )
 
         def _D_cobalt_lithium_calculation(self, x, z):
-            param_cobalt_lithium = [-6.98e-07, 2.27e-08, -4.4e-09]
+            if value(self.membrane_fixed_charge) == 0:
+                # these params assume chi=0 mM and lithium & cobalt mem_conc range 50-80 mM
+                self.diffusion_param_cobalt_lithium = [-5.24e-07, 2.49e-09, -2.50e-09]
+            if value(self.membrane_fixed_charge) == -140:
+                # these params assume chi=-140 mM and lithium mem_conc range 50-80 mM & cobalt mem_conc range 80-110 mM
+                self.diffusion_param_cobalt_lithium = [-1.32e-06, 4.72e-09, 1.47e-09]
```
dallan-keylogic, line 821: Same comment as above.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -746,38 +846,110 @@ def _D_cobalt_lithium_calculation(self, x, z):
         )
 
         def _D_cobalt_cobalt_calculation(self, x, z):
-            params_cobalt_cobalt = [-4.04e-06, 2.17e-07, -4.2e-08]
+            if value(self.membrane_fixed_charge) == 0:
+                # these params assume chi=0 mM and lithium & cobalt mem_conc range 50-80 mM
+                self.diffusion_params_cobalt_cobalt = [-4.00e-06, 6.45e-09, -6.48e-09]
+            if value(self.membrane_fixed_charge) == -140:
+                # these params assume chi=-140 mM and lithium mem_conc range 50-80 mM & cobalt mem_conc range 80-110 mM
+                self.diffusion_params_cobalt_cobalt = [-6.05e-06, 1.22e-08, 3.81e-09]
```
dallan-keylogic, line 854: Same comment as above.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -746,38 +846,110 @@ def _D_cobalt_lithium_calculation(self, x, z):
...
             self.x_bar, self.z_bar, rule=_D_cobalt_cobalt_calculation
         )
 
+        def _convection_coefficient_lithium_calculation(self, x, z):
+            if value(self.membrane_fixed_charge) == 0:
+                # these params assume chi=0 mM and lithium & cobalt mem_conc range 50-80 mM
+                self.convection_params_lithium = [1, 0, 0]
+            if value(self.membrane_fixed_charge) == -140:
+                # these params assume chi=-140 mM and lithium mem_conc range 50-80 mM & cobalt mem_conc range 80-110 mM
+                self.convection_params_lithium = [0.260, 0.00139, 0.00314]
```
dallan-keylogic, line 887: Same comment as above

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -746,38 +846,110 @@ def _D_cobalt_lithium_calculation(self, x, z):
...
+            self.x_bar, self.z_bar, rule=_convection_coefficient_lithium_calculation
+        )
+
+        def _convection_coefficient_cobalt_calculation(self, x, z):
+            if value(self.membrane_fixed_charge) == 0:
+                # these params assume chi=0 mM and lithium & cobalt mem_conc range 50-80 mM
+                self.convection_params_cobalt = [1, 0, 0]
+            if value(self.membrane_fixed_charge) == -140:
+                # these params assume chi=-140 mM and lithium mem_conc range 50-80 mM & cobalt mem_conc range 80-110 mM
+                self.convection_params_cobalt = [0.0860, 0.00198, 0.00448]
```
dallan-keylogic, line 914: Same comment as above.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -1063,70 +1245,55 @@ def add_scaling_factors(self):
...
-                self.scaling_factor[self.D_lithium_cobalt_calculation[x, z]] = 1e5
-                self.scaling_factor[self.D_cobalt_lithium_calculation[x, z]] = 1e5
-                self.scaling_factor[self.D_cobalt_cobalt_calculation[x, z]] = 1e6
-
-                if z != 0:
-                    self.scaling_factor[self.d_membrane_conc_mass_lithium_dz[x, z]] = (
-                        1e4
-                    )
-                    self.scaling_factor[self.d_membrane_conc_mass_cobalt_dz[x, z]] = 1e2
+        constraint_autoscale_large_jac(self)
```
dallan-keylogic, line 1266: This is all going to have to get overhauled to use the new scaler classes soon.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
```diff
@@ -36,6 +36,86 @@
...
+    assert m.fs.unit.diffusion_params_lithium_lithium == [
+        -4.07e-06,
+        -3.99e-09,
+        4.01e-09,
+    ]
+    assert m.fs.unit.diffusion_params_lithium_cobalt == [-9.62e-07, -1.03e-08, 1.04e-08]
+    assert m.fs.unit.diffusion_param_cobalt_lithium == [-5.24e-07, 2.49e-09, -2.50e-09]
+    assert m.fs.unit.diffusion_params_cobalt_cobalt == [-4.00e-06, 6.45e-09, -6.48e-09]
+    assert m.fs.unit.convection_params_lithium == [1, 0, 0]
+    assert m.fs.unit.convection_params_cobalt == [1, 0, 0]
```
dallan-keylogic, line 62: Instead of just being lists that live on the unit model, these should probably be indexed `Vars` because I'm sure we'll want to do parameter estimation sometime in the future.

dallan-keylogic (changes requested):

src/prommis/nanofiltration/diafiltration_two_salt.py
mollydougher: Meaning they can be chained together (retentate-staged or permeate-staged) to create a multi-stage diafiltration cascade.

src/prommis/nanofiltration/diafiltration_two_salt.py
mollydougher: They are derived based on equilibrium relationships.

bpaul4 (commented):
Just a couple more comments, I think everything looks good.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -465,38 +473,95 @@ def initialize_diafiltrate_conc_mol_comp(m, t, j):
             doc="Mole concentration of solutes in the diafiltrate",
         )
 
-        # add variables dependent on x_bar
+        # add parameters for the linearized relationships to calculate the diffusion and convection coefficients
```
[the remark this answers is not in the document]
bpaul4: A bit more commentary or explanation is probably sufficient here.

src/prommis/nanofiltration/tests/test_diafiltration_two_salt.py
```diff
@@ -613,47 +693,47 @@ def test_solve(self, diafiltration_two_salt):
         test_dict = {
```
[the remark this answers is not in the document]
mollydougher: Yes. The retentate (decrease) and permeate (increase) flow rates change due to the increased parameter values for the applied pressure and membrane area. The concentrations see significant changes mainly due to the conversion from a mass basis to a mole basis, and there are small changes with the addition of partitioning.

src/prommis/nanofiltration/diafiltration_flowsheet_two_salt.py
```diff
@@ -74,11 +75,62 @@ def main():
...
+                ):
+                    raise ValueError(
+                        "Membrane concentration for cobalt ("
+                        f"{value(m.fs.membrane.membrane_conc_mol_cobalt[x, z])} mM at "
+                        f"x={x * value(m.fs.membrane.total_module_length)} m and "
+                        f"z={z * value(m.fs.membrane.total_membrane_thickness)} m) is outside "
+                        "of the valid range for the diffusion coefficient approximations "
+                        "(50-80 mM). The linearized approximation should be re-calculated."
+                    )
+
```
dallan-keylogic, line 129: There's an old and little-used API for validation like this: the `model_check` method on the [unit model block](https://github.com/IDAES/idaes-pse/blob/7d19f17d8bbf3c7790ddef827751a5da3a12648c/idaes/core/base/unit_model.py#L118). It would probably be good to implement these checks by implementing a `model_check` method on your unit model.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -10,42 +10,44 @@
...
+The Two-Salt Diafiltration unit model requires a property package that provides the valency (:math:`z_i`), reflection coefficient (:math:`\sigma_i`), partition coefficient (:math:`H_i`), and number of dissolved species (:math:`n_i`) for each ion :math:`i` in solution. When used in a flowsheet, the user can provide separate property packages for the feed and product streams.
 
-Additionally, there are two required arguments, ``NFEx`` and ``NFEz``, to specfiy the desired number of finite elements across the width and thickness of the membrane, respectively.
+There are two required arguments, ``NFE_module_length`` and ``NFE_membrane_thickness``, to specify the desired number of finite elements across the width and thickness of the membrane, respectively. There is one optional argument, ``charged_membrane``, which is a Boolean (default = ``True``) to specify if the membrane has a fixed surface charge.
```
dallan-keylogic, line 20: Does the model structure change depending on whether the membrane has a fixed surface charge or is uncharged?

dallan-keylogic: It seems like the big thing that changes is which set of parameters is loaded in. Instead of using a `Bool` here, consider using an `Enum` like is done [in the heat exchanger.](https://github.com/IDAES/idaes-pse/blob/7d19f17d8bbf3c7790ddef827751a5da3a12648c/idaes/models/unit_models/heat_exchanger.py#L59) That allows you to naturally extend the system if you want to try, for example, a third level of charge.

dallan-keylogic (commented):
Some comments.

src/prommis/nanofiltration/diafiltration_two_salt.py
mollydougher: Similarly, I think this comment can be moved to the back burner, as the analytical equations take the fixed membrane charge as an input and there are no longer different parameter sets to load in.

bpaul4 (approved):
Everything looks great to me and all my concerns have been addressed.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -351,243 +364,319 @@ def add_mutable_parameters(self):
...
+            initialize=0.6,
+            units=units.dimensionless,
+            doc="Concentration-dependent convection coefficient for lithium",
+        )
+        self.convection_coefficient_cobalt = Var(
+            self.dimensionless_module_length,
+            self.dimensionless_membrane_thickness,
+            initialize=0.5,
+            units=units.dimensionless,
+            doc="Concentration-dependent convection coefficient for cobalt",
```
dallan-keylogic: Make these indexed by time, module length, membrane thickness, and component.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -351,243 +364,319 @@ def add_mutable_parameters(self):
...
             doc="Lithium concentration gradient wrt z in the membrane",
         )
-        self.d_membrane_conc_mass_cobalt_dz = DerivativeVar(
-            self.membrane_conc_mass_cobalt,
-            wrt=self.z_bar,
-            units=units.kg / units.m**3,
+        self.d_membrane_conc_mol_cobalt_dz = DerivativeVar(
+            self.membrane_conc_mol_cobalt,
+            wrt=self.dimensionless_membrane_thickness,
+            units=units.mol / units.m**3,  # mM
```
dallan-keylogic: These should also be indexed by component.

dallan-keylogic (commented):
Overall, this looks pretty good, but there are a few changes that would be nice.

1) There are a lot of cases in which variables and constraints have the component they apply to in their names. It would be better if those components were instead indexed by those components instead---that makes the model structure easier to inspect, as well as makes it easier to apply the model to other components.

2) Similarly, a lot of variables and constraints should be able to be indexed by time, but it looks like time indexing is completely ignored inside the unit model.

3) Most of the relationships in the `fix_boundary_values` method should be `Constraints`. (The main exception is deactivating the chloride concentration gradient and fixing the discretization equations).

dallan-keylogic (commented):
Just a few more minor changes. I'd be okay with merging it as-is if there is a time crunch.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -351,706 +347,883 @@ def add_mutable_parameters(self):
...
         )
 
-    def discretize_model(self):
-        discretizer = TransformationFactory("dae.finite_difference")
-        discretizer.apply_to(
-            self, wrt=self.x_bar, nfe=self.config.NFEx, scheme="BACKWARD"
+        def _permeate_conc_mol_comp_boundary_condition(blk, t, j):
+            return (
+                blk.permeate_conc_mol_comp[t, 0, j]
+                == value(self.numerical_zero_tolerance) * units.mol / units.m**3
```
dallan-keylogic:
```suggestion
                == self.numerical_zero_tolerance * units.mol / units.m**3
```

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -351,706 +347,883 @@ def add_mutable_parameters(self):
...
-        self.retentate_conc_mass_comp[0, 0, "Co"].fix(
-            (
-                self.feed_flow_volume[0] * self.feed_conc_mass_comp[0, "Co"]
-                + self.diafiltrate_flow_volume[0]
-                * self.diafiltrate_conc_mass_comp[0, "Co"]
+
+        def _molar_ion_flux_boundary_condition(blk, t, j):
+            return (
+                blk.molar_ion_flux[t, 0, j]
+                == value(self.numerical_zero_tolerance)
```
dallan-keylogic:
```suggestion
                == self.numerical_zero_tolerance
```

dallan-keylogic (approved):
Just a few suggestions. When using these numerical zero `Params`, it's useful to be able to tweak them---moving them further away from zero can sometimes help convergence.

src/prommis/nanofiltration/diafiltration_two_salt.py
```diff
@@ -351,706 +347,818 @@ def add_mutable_parameters(self):
...
-        )
-        self.d_retentate_conc_mass_comp_dx[0, 0, "Co"].fix(
-            value(self.numerical_zero_tolerance)
-        )
-        self.d_retentate_flow_volume_dx[0, 0].fix(value(self.numerical_zero_tolerance))
+            for j in self.solutes:
+                self.permeate_conc_mol_comp[t, 0, j].fix(
+                    value(self.numerical_zero_tolerance)
+                )
+                self.molar_ion_flux[t, 0, j].fix(value(self.numerical_zero_tolerance))
```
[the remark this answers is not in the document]
mollydougher: @dallan-keylogic The diafiltrate is essentially an additional feed stream that we supply to the membrane, so this enters on the retentate-side of the membrane with the feed stream. Anything that fluxes through the membrane will enter the permeate-side of the membrane. We assume that the retentate at x=0 consists of only the feed streams (feed + diafiltrate). There is no flux through the membrane only at x=0, so the membrane concentrations and permeate concentrations are 0 at x=0.

Issue #152, opened 2025-05-20
UKy REE Recovery


MarcusHolly:
In #138, it was determined that the three acid (HCl) feed flows had unrealistic flow rates, so they were increased by orders of magnitude. While this helped the system to converge to an optimal solution, the total REE recovery dropped from ~20% to around ~0.2%. Until we get better data, we should adjust the operating conditions of the flowsheet so that we get higher REE recoveries. One variable that has been shown to affect `m.fs.sl_sep2.liquid_recovery.fix(0.9)`

ksbeattie:
@MarcusHolly, any update on this?

PR #204, opened 2026-01-14
Improved method to compare solution to expected values

dallan-keylogic:
## Summary/Motivation:
As a part of addressing #152, we're going to be making major changes to the UKy flowsheet. As a result, the tests in which we check the solution of the UKy flowsheet against cached values are going to be changed (probably multiple times). I got tired of having to change those values one assertion statement at a time, so prepared a function that iterates over a bunch of values at once, determines which values do and do not match, then creates a single report with all the variables that don't have their expected values.

This function is going to be added to IDAES by the next release, but we need to add it temporarily to PrOMMiS because I don't want to change the version of IDAES that PrOMMiS depends on before we're almost ready to cut a release.

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.

### DIFF
--- a/src/prommis/uky/tests/test_uky_flowsheet.py
+++ b/src/prommis/uky/tests/test_uky_flowsheet.py
@@ -8,7 +8,6 @@
 Tests for UKy flowsheet.
 
 """
-import textwrap
 import pytest
 
 from pyomo.environ import (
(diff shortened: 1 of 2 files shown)


bpaul4 (approved):
This is great and very useful! Will the test file structure be reverted back to a long list of assert statements later, or will this be the recommended format going forwards for long lists of results in test files?

MarcusHolly (approved):
This looks great, thanks for putting this together Doug!

sufikaur (approved):

PR #205, opened 2026-01-19
Leaching changes

dallan-keylogic:
## Summary/Motivation:
Adds scaling for the leaching train, as well as the leach solution and leach solids property packages. The leach solution and solids property packages also had some properties added or existing properties were converted to on-demand properties. The leach train also gained a constraint relating the rate of outflow of the solid and liquid streams to their phase holdups.

These changes broke the dynamic leaching model, but I determined that there was no point in fixing the old implementation when I have a local version that applies DAE index reduction in a systematic way. I'll open another PR with that model after this one is merged.

These changes also revealed that the current implementation of the UKy flowsheet is extremely fragile. The material balance constraints are badly scaled, so significant amounts of material can be created out of thin air while still satisfying the convergence criteria of IPOPT. The solution is scaling the entire flowsheet, which will have to wait until after I merge in scaling for solvent extraction.

- [x] Documentation
- [x] Tests
- [x] Diagnostic tests for models

### Legal Acknowledgement

   employer.


src/prommis/properties/coal_refuse_properties.py
```diff
@@ -164,33 +250,44 @@ def build(self):
...
-                j
-            ] * b.mass_frac_comp["inerts"] == b.mass_frac_comp[
+                return b.conversion_comp[j] == b.params.eps
+            # Fudge the initial mass fraction of non-inerts
+            # so that the initial conversion isn't exactly on its bounds
+            return (1 - b.conversion_comp[j]) * (
+                1 + b.params.eps
+            ) * b.params.mass_frac_comp_initial[j] * b.mass_frac_comp[
+                "inerts"
+            ] == b.mass_frac_comp[
```
MarcusHolly: Is this the convention for handling the value for initial conversion? If not, why is this necessary for UKy?

MarcusHolly (approved):
Minor comments, but overall LGTM

src/prommis/properties/coal_refuse_properties.py
dallan-keylogic: Is there another place where we're using conversion as a variable on the property package?

Previously, the `conversion` variable (which I renamed to `conversion_comp` in this PR) had no lower bound. However, when troubleshooting issues in the leaching dynamics, I encountered "large" (above 1e-2)  negative values of `conversion`, so I added a lower bound of zero. That had the unpleasant side effect of that lower bound being active in the solids feed (the conversion was 0 because nothing had been processed) and it resulted in `assert_no_numerical_warnings` failing in a test.

I added `eps` as a fudge factor to make sure that the initial conversion was not exactly zero while (hopefully) not significantly affecting further results. For non-inert components, we're effectively treating the initial mass fraction as one part in `1e-8` larger for the purpose of calculating conversion, whereas for inert components the conversion is fixed at `1e-8`. 

Adding this fudge factor isn't the reason why the results in the UKy flowsheet changed---even when I set `eps=0`, the results were wildly different.

src/prommis/properties/coal_refuse_properties.py
```diff
@@ -164,33 +250,44 @@ def build(self):
         self.mass_frac_comp = Var(
             self.params.component_list,
             units=units.kg / units.kg,
-            bounds=(1e-8, None),
+            bounds=(1e-12, 1.01),
```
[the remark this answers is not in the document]
dallan-keylogic: Conversion has an upper bound of 0.9999 because the equation for conversion may become degenerate when conversion equals exactly 1 (it has a term of `1 - b.conversion_comp[j]`). I don't know of any similar concern for mass fraction, so I'd like to avoid cutting off any feasible points.

src/prommis/properties/coal_refuse_properties.py
dallan-keylogic: We can examine that property package for issues at some point.

bpaul4 (commented):
Just a few comments.

bpaul4 (approved):

PR #214, opened 2026-02-06
Generalize multi-component diafiltration unit model

mollydougher:
## Summary/Motivation:
The multi-component diafiltration unit model currently only supports the separation of a lithium chloride and cobalt chloride solutions. By generalizing the model to any number of cations and providing additional cations within the property package, we can apply this high-fidelity unit model to more complex feed streams.

- Rename the multi-salt diafiltration files from `..._two_salt` to `multi_component_...` to better describe the model capabilities.
- Generalize the unit model to any cation (the user provides a list of the cations in solution).
- Expand the property packages to more components and solutions.
- Support the generation of the multi-component diafiltration unit model with 1, 2, or 3 salts comprised of +1, +2, and +3 ions.
- Updated the default value for the fixed membrane charge based on experimental data.

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/nanofiltration/multi_component_diafiltration.py
```diff
@@ -361,8 +368,15 @@ def add_variables(self):
...
-        self.cations = Set(initialize=["Li", "Co"])
+        if self.config.num_salts == 1:
+            self.solutes = Set(initialize=["cation_1", "anion"])
+            self.cations = Set(initialize=["cation_1"])
+        if self.config.num_salts == 2:
+            self.solutes = Set(initialize=["cation_1", "cation_2", "anion"])
+            self.cations = Set(initialize=["cation_1", "cation_2"])
+        if self.config.num_salts == 3:
+            self.solutes = Set(initialize=["cation_1", "cation_2", "cation_3", "anion"])
+            self.cations = Set(initialize=["cation_1", "cation_2", "cation_3"])
```
dallan-keylogic: Instead of making `num_salts` the config option, make `cation_list` a config option, then create a Pyomo set initialized by that list. That way you don't have to remember what `cation_1` and `cation_2` are when plotting your results.

You should also make `anion_list` a config option. I don't know if you're going to look at multi-anion systems in the future, but for now you can just check that the length of the provided list is equal to 1.

(Also, the user should always provide Python lists, not Python sets, for the config. Python sets are unordered, which means there's no set order when iterating over them, which can lead to obscure errors.)

dallan-keylogic (commented):
Just a few minor changes.

dallan-keylogic:
Since you've created an initialization method, you should probably use the new IDAES initialization API. The way we do initialization now is to create an `Initializer` class, then attach it to the `UnitModelBlock` as the `default_initializer` attribute. You can look at, e.g., the [leaching unit model](https://github.com/prommis/prommis/blob/aaec8345274867743b16db957685e92625da56c0/src/prommis/leaching/leach_train.py#L285) to see how that's done.

How I would do it is to create a subclass of the [block triangularization initializer](https://github.com/IDAES/idaes-pse/blob/da90c69ce114d68be3294e271088ca196d3210c2/idaes/core/initialization/block_triangularization.py#L34). Then you would overwrite the `initialization_routine` method with your `initialize_streams` method. Then, after the streams are initialized, call `super().initialization_routine` to call block triangularization on your model. Since the unit model is cocurrent, you should end up with relatively small blocks (countercurrent unit models typically end up with one enormous block containing most of the variables and constraints in the model).

Then to call the initializer method, you would use
```
init_obj = m.fs.unit.default_initializer()
init_obj.initialize(m.fs.unit)
```

dallan-keylogic (approved):
LGTM

tristantc (commented):
LGTM. Just a minor point: wouldn't it be more consistent to standardize how species are named across the models, using `Li_+` instead of `lithium` or `Li`?

tristantc (approved):
LGTM

PR #215, opened 2026-02-09
Costing Hydrogen Decrepitation Furnace

Akintomiwaojo:
## Summary/Motivation:
This work estimates the cost of a hydrogen decrepitation furnace using a bottom-up approach.

- Add a file to estimate the cost of a hydrogen decrepitation furnace.
- Add the associated test files.

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


MarcusHolly (approved):

src/prommis/hydrogen_decrepitation/costing/cost_hydrogen_decrepitation_furnace.py
```diff
@@ -0,0 +1,410 @@
...
+
+_log = idaeslog.getLogger(__name__)
+
+
+def custom_REE_plant_currency_units():
+    """
+    Define conversion rates for US Dollars based on CE Index.
+    """
+    register_idaes_currency_units()
+    if "USD_Jan_2024" in pyunits._pint_registry:  # pylint: disable=protected-access
```
dallan-keylogic, line 36: @blnicho , does `pyomo.environ.units` have a public API for checking this?

src/prommis/hydrogen_decrepitation/costing/cost_hydrogen_decrepitation_furnace.py
```diff
@@ -0,0 +1,410 @@
...
+        """
+        Cost Estimation of an Hydrogen Decrepitation Furnace
+
+        This method calculates the capital cost of a hydrogen decreptitation furnace.
+
+        Args:
+        blk: A unit-level costing block where costing variables and
+             constraints can be added to.
+        heating_mode: mode of operation of the heating system. 0 - electricity, 1 - gas-fired
+        """
```
dallan-keylogic, line 178: Because `blk` is never going to be an instance of `HydrogenDecrepitationCosting` or `HydrogenDecrepitationCostingData`, it doesn't belong as part of the flowsheet costing class. It should be turned into a separate function.

dallan-keylogic (changes requested):
Unfortunately, it looks like some of my comments from last time around didn't get posted. 

The main point of feedback is that `cost_hydrogen_decrepitation_furnace` doesn't really belong on the class to which it's attached (unless I'm missing some subtlety about how `custom_REE_plant_currency_units` is getting called in the header). If it gets called by an instance of a `HydrogenDecrepitationCosting` block, it will raise an exception. It should be a separate function.

src/prommis/hydrogen_decrepitation/costing/cost_hydrogen_decrepitation_furnace.py
bpaul4: Would you prefer we move it inside the unit model so we can call it as `m.fs.hydrogen_decrepitation_furnace.add_costing()` and force it to use the main PrOMMiS costing attributes, class, hard code things like currency units?

Or keep it as a separate model like https://github.com/prommis/prommis/tree/main/src/prommis/nanofiltration/costing?

dallan-keylogic: It should be on a subclass of a `UnitModelCostingBlock`---that's going to be what `blk` ends up being when the function is called.

src/prommis/hydrogen_decrepitation/costing/cost_hydrogen_decrepitation_furnace.py
```diff
@@ -0,0 +1,410 @@
...
+        )
+        self.engineering_and_drafting.set_value(1000 * pyunits.USD_Jan_2024)
+
+        self.CE_index_year = Param(
+            initialize="Jan_2024",
+            mutable=True,
+            doc="Base year to convert costing values to, e.g. 'Jan_2024'",
+        )
+
+    def cost_hydrogen_decrepitation_furnace(
```
dallan-keylogic, line 165:
```suggestion
    @staticmethod
    def cost_hydrogen_decrepitation_furnace(
```

After talking with @bpaul4 and looking at some of the other costing code, the best solution is to make this function a static method. Because it's a static method, a `HydrogenDecrepitationCosting` that calls the method will not get passed to the method automatically as the first argument. This change clears up the mismatch between the fact that this method lives on the `HydrogenDecrepitationCosting` block but is not ever called as a method for that block.

After this small change is made, it should be ready to merge.

dallan-keylogic (changes requested):

dallan-keylogic (approved):
LGTM

PR #220, opened 2026-03-02
Updating UKy Acid Feed Conditions

MarcusHolly:
## Addresses Issue: 
The first item in https://github.com/prommis/prommis/issues/152

## Summary/Motivation:
There was a mistake in REESim where the strength (wt%) of HCl is cited as 0.037%; however, this should actually be 3.7%, which aligns with the 1M HCl that we have confirmed to be correct.

- Updates the acid feed compositions and flow rates
- Updates the UKy flowsheet test file

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/uky/uky_flowsheet.py
```diff
@@ -920,10 +920,11 @@ def set_operating_conditions(m):
     m.fs.rougher_org_make_up.conc_mass_comp[0, "DEHPA"].fix(dehpa_conc)
     m.fs.rougher_org_make_up.conc_mass_comp[0, "Kerosene"].fix(kerosene_conc)
 
-    m.fs.acid_feed1.flow_vol.fix(90)  # Increased by 1000x from REESim
+    # Assumes an HCl weight percent of 3.7%
+    m.fs.acid_feed1.flow_vol.fix(0.09)
     m.fs.acid_feed1.conc_mass_comp[0, "H2O"].fix(1000000)
```
MarcusHolly: @dallan-keylogic Perhaps we should follow up with the UKy team about this acid feed flow rate. As a recap, the REESim file suggested that there is 16.05 kg/hr of Fe in a aqueous stream of 0.09 kg/hr (0.09 L/hr), which does not make much physical sense. For now, I've reverted back to the 0.09 L/hr value instead of our arbitrary value of 90 L/hr, since it does help boost the REE recovery. If we change the HCl weight percent w/o adjusting the acid feed flow rates, we actually lower the recovery even more

src/prommis/properties/hcl_stripping_properties.py
```diff
@@ -57,10 +57,10 @@ class HClStrippingPropertiesScaler(CustomScalerBase):
     DEFAULT_SCALING_FACTORS = {
         "flow_vol": DefaultScalingRecommendation.userInputRequired,
         "conc_mass_comp[H2O]": 1e-6,
-        "conc_mass_comp[H]": 1e-1,
+        "conc_mass_comp[H]": 1e-3,
         "conc_mass_comp[SO4]": 1e-2,
         "conc_mass_comp[HSO4]": 1e-3,
-        "conc_mass_comp[Cl]": 1e-3,
+        "conc_mass_comp[Cl]": 1e-5,
```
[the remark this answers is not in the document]
MarcusHolly: Yes, I'll make this change

src/prommis/uky/uky_flowsheet.py
```diff
@@ -2682,13 +2688,21 @@ def volume_scaling_constraint(c, k):
...
+
+    # Deactivate all constraints
+    for c in m.fs.costing.component_data_objects(Constraint):
+        c.deactivate()
+
+    # Activate the costing constraints
+    for c in m.fs.scaling_constraints.values():
+        c.activate()
+
+    solve_strongly_connected_components(m.fs.costing, solver=get_solver("ipopt_v2"))
```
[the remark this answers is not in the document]
dallan-keylogic: Yes, you're activating the constraints on `m.fs.scaling_constraints`, but they were never deactivated to begin with. The issue here is that the constraints added by costing are spread out on a whole bunch of different blocks. You could import `create_subsystem_block` from `pyomo.util.subsystems`, then add all of the costing variables and constraints onto a temporary block, then solve that block using `solve_strongly_connected_components`. So that would be all the variables and constraints from `m.fs.scaling_constraints`, the variables and constraints from `m.fs.costing`, and all the other costing variables and constraints that exist on the `costing` subblocks on all the unit models.

src/prommis/uky/uky_flowsheet.py
```diff
@@ -269,14 +274,6 @@ def main():
...
 
-    auto = AutoScaler()
-    auto.scale_variables_by_magnitude(m)
-    auto.scale_constraints_by_jacobian_norm(m)
-
-    QGESSCostingData.costing_initialization(m.fs.costing)
-    QGESSCostingData.initialize_fixed_OM_costs(m.fs.costing)
-    QGESSCostingData.initialize_variable_OM_costs(m.fs.costing)
-
     solve_system(m, tee=True)
```
dallan-keylogic: Are we not initializing costing as part of `main` anymore? Will it successfully solve without initialization?

MarcusHolly: Initialization is being completely handled in initialize_costing now (i.e. these steps are no longer necessary). In fact, the system was only able to successfully solve after I removed these steps, so there was probably some negative interference between this initialization and initialize_costing

dallan-keylogic: Ah, I forgot we were already calling that method; it wasn't something you added just now.

So, even without the extra initialization, the autoscaling makes it fail? I thought that costing would be one place where the autoscaling would work.

dallan-keylogic (approved):
LGTM

bpaul4 (approved):
LGTM, with a note that we should update the examples as needed when the flowsheet itself is finalized.

PR #243, opened 2026-05-07
Crusher add equipment

Lingyan90:
## Addresses Issue: 
Expand the existing crusher unit model to allow for more detailed stage and equipment specific crushing

## Summary/Motivation:
This expansion enlarges the equipment bank to include a broader set of industry-relevant equipment. It also enables future flowsheet development with more detailed equipment selection, including crusher-stage selection based on applicable particle-size ranges.

- added crushing stage category: primary, secondary, and tertiary
- added crusher stage and equipment selection criteria
- added configuration for crusher stage, equipment and size limits
- add debug note when provided stage category and equipment didn't match
- added constraint crushing direction
- added applicability check to inform whether the current product P80 is consistent with the configured stage classification interval
- updated model build in test file
- added a series of test to test different scenarios of crusher stage and equipment selection

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


src/prommis/solid_handling/crusher.py
```diff
@@ -10,57 +10,96 @@
...
+80% passing the mesh at :math:`t` time.
 
 Expressions
 -----------
 
-Crusher includes two expressions to calculate the size of particles that has 80% passing the mesh for both feed and product particles.
+Crusher includes two expressions to calculate the size of particles that has 80%
+passing the mesh for both feed and product particles.
 
 .. math:: P_{t, feed, 80} = \frac{S_{t, in, median}}{unit} * \left(-\log(1 - 0.8)\right)^{\frac{SW_{t, in}}{2}}
```
sk-netl, line 76: I'm struggling with this, I think a little more detail or a citation/reference on this relationship would help.  Is the link between particle size and median based on the Gaudin-Schumann distribution?

Lingyan90: it is a Rosin-Rammler particle size distribution: https://doi.org/10.1016/0304-3967(80)90007-4. line 75 show the back calculation of the particle size that allow 80% of the particles passing the mesh. I'll add the citation in an updated push.

src/prommis/solid_handling/crusher.py
```diff
@@ -75,6 +114,123 @@
 
 _log = idaeslog.getLogger(__name__)
 
+STAGE_P80_EPS = 1e-6 * pyunits.cm
+CRUSHER_STAGE_ORDER = ("primary", "secondary", "tertiary")
```
Daison2102: `CRUSHER_STAGE_ORDER` is defined here but never used anywhere in the file. Can we remove it?

src/prommis/solid_handling/crusher.py
```diff
@@ -75,6 +114,123 @@
...
+    if _product_p80_in_stage_range("primary", product_p80_native, native_units):
+        return "primary", True
+
+    if _product_p80_in_stage_range("secondary", product_p80_native, native_units):
+        return "secondary", True
+
+    if _product_p80_in_stage_range("tertiary", product_p80_native, native_units):
+        return "tertiary", True
+
+    return "tertiary", False
```
Daison2102, line 233: `_recommend_stage_from_product_p80_native` reuses `_product_p80_in_stage_range`, which clips the secondary and tertiary upper bounds by `eps` (`1e-6 cm ≈ 10 nm`) to keep the Pyomo algebraic constraints non-overlapping. That's fine for the solver, but it leaves tiny gaps in the recommendation logic: a P80 of e.g. `9.9999991 cm` is above secondary's clipped upper bound but below primary's lower bound, so it falls through all three checks and hits the `("tertiary", False)` fallback - which `check_applicability()` reports as "finer than the minimum modeled crushing range (0.5 cm)". Wrong message for a value near 10 cm.

The `eps` clipping belongs only in the algebraic constraints. The recommendation function should instead use simple waterfall comparisons against the lower bounds (>= 10 cm -> primary, >= 2 cm -> secondary, >= 0.5 cm -> tertiary), so `False` means exactly one thing: below 0.5 cm. The gap is physically negligible (10 nm), but the fix is small and makes the logic unambiguous.

src/prommis/solid_handling/crusher.py
```diff
@@ -122,6 +278,82 @@ class CrusherData(UnitModelBlockData):
...
+- roll2
+- short_head_cone
+- hammer_mill
+""",
+        ),
+    )
+    CONFIG.declare(
+        "enforce_stage_size_limits",
+        ConfigValue(
+            default=True,
```
Daison2102: Two new defaults combine to create a silent behavior change. `crusher_stage` defaults to `None`, but that resolves to `"secondary"` at build time, while `enforce_stage_size_limits` defaults to `True`. As a result, a plain `Crusher(property_package=...)` now adds hard algebraic constraints requiring product P80 to be in the secondary crusher range, even when the user never selected a stage.

Before this PR, the existing test case used `particle_size_median = 58 µm`, giving product P80 of about 83 µm. The new secondary lower bound is 2 cm = 20000 µm, so the same model specification is now outside the default feasible region.

If hard stage enforcement is intended, I think it should be explicit and documented as a breaking behavior change. Otherwise, can we either default `enforce_stage_size_limits=False`, or only enforce stage bounds when the user explicitly supplies `crusher_stage` or `crusher_equipment`?

The failure mode is the hard part: if the model becomes infeasible because of the new stage bounds, the user gets a solver convergence error with no pointer to the new constraint.

src/prommis/solid_handling/crusher.py
```diff
@@ -122,6 +278,82 @@ class CrusherData(UnitModelBlockData):
...
+            domain=In([True, False]),
```
Daison2102: can we use `Bool` from `pyomo.common.config` instead of `In([True, False])`?

src/prommis/solid_handling/crusher.py
```diff
@@ -122,6 +278,82 @@ class CrusherData(UnitModelBlockData):
...
+        ),
+    )
+
+    def _resolve_stage_selection(self):
+        selected_stage = self.config.crusher_stage
+        selected_equipment = self.config.crusher_equipment
+
+        # Return values are:
+        # (resolved_stage, resolved_equipment, selection_basis)
+        if selected_stage is None and selected_equipment is None:
```
Daison2102, line 339: Can you point to the specific source/table for the product-P80 bounds, equipment-to-stage mapping, and the choice to default an unspecified crusher to `secondary`? I can see the citation in the module docstring, but because these values now affect feasibility, it would help to know whether they are intended as hard model-validity limits or as rough applicability guidance.

src/prommis/solid_handling/crusher.py
```diff
@@ -122,6 +278,82 @@ class CrusherData(UnitModelBlockData):
...
+            return "secondary", None, "default"
+
+        if selected_stage is not None and selected_equipment is None:
+            return selected_stage, None, "stage"
+
+        selected_equipment = selected_equipment.strip().lower()
```
Daison2102: `selected_equipment.strip().lower()` is dead code. The `crusher_equipment` config domain uses `In([None] + list(CRUSHER_EQUIPMENT_TO_STAGE.keys()))`, which rejects any string that does not exactly match the list - such as `"  cone  "` or `"Cone"` - before this method is ever called. The strip/lower always receives an already-exact string and has no effect. It could mislead a reader into thinking non-canonical inputs are supported when they are not.

src/prommis/solid_handling/crusher.py
```diff
@@ -180,6 +412,36 @@ def build(self):
         """
 
         sunit = self.properties_in[tref].particle_size_median.get_units()
+        stage_name, equipment_name, selection_basis = self._resolve_stage_selection()
+        stage_spec = CRUSHER_STAGE_DATA[stage_name]
+
+        self._crusher_stage = stage_name
```
Daison2102, line 420: The resolved stage info is stored as private attributes (`_crusher_stage`, `_crusher_equipment`, etc.) but the tests read them directly to verify stage selection. Since the resolved stage now affects model feasibility, users will want to inspect it too. Private attributes carry no stability guarantee - the name could change without it being considered a breaking change. Can we expose `crusher_stage` and `crusher_equipment` as proper public properties?

src/prommis/solid_handling/crusher.py
```diff
@@ -180,6 +412,36 @@ def build(self):
...
+        self._applicable_product_p80_upper = _range_bound_to_native_units(
+            stage_spec["applicable_product_p80_upper"], sunit
+        )
+        self._stage_product_p80_eps = _stage_eps_to_native_units(sunit)
+
+        default_product_p80 = _size_to_native_units(
+            stage_spec["default_product_p80"], sunit
+        )
+
+        for t in self.flowsheet().time:
```
Daison2102, line 442: Is the build-time reset of the outlet median particle-size initial value intended to be part of the public behavior? This changes the property block's initialized value based on crusher stage before the user fixes states. I am not sure whether that is meant as an initialization aid or whether users should provide the outlet particle-size initial guess themselves.

src/prommis/solid_handling/crusher.py
```diff
@@ -206,6 +468,163 @@ def crush_work_eq(self, t):
                 * (1 / (self.prod_p80[t]) ** 0.5 - 1 / (self.feed_p80[t]) ** 0.5)
             )
 
+        @self.Constraint(
```
Daison2102, line 477: When all particle size state variables are fixed (the standard test and flowsheet setup), `crushing_direction_constraint`, `product_p80_lower_bound`, and `product_p80_upper_bound` all have zero Jacobian rows, which IDAES `DiagnosticsToolbox.assert_no_numerical_warnings()` flags as degenerate. The test works around this by deactivating `crushing_direction_constraint` and setting `enforce_stage_size_limits=False`. Is the intended usage of these constraints that the outlet particle size must remain free (not fixed by the user), with the solver determining it subject to the bounds? Or should users always fix all particle sizes and only use `check_applicability()` as a post-hoc check?

src/prommis/solid_handling/tests/test_crusher.py
```diff
@@ -36,65 +39,167 @@
...
-            0.1
-        )  # set mass frac value. There are 13 component in property package
-        m.fs.unit.properties_in[0].flow_mass.fix(2000)  # kg/hr
-        m.fs.unit.properties_in[0].particle_size_median.fix(80)  # micrometer
-        m.fs.unit.properties_in[0].particle_size_width.fix(1.5)  # dimensionless
-        m.fs.unit.properties_out[0].particle_size_median.fix(58)  # micrometer
-        m.fs.unit.properties_out[0].particle_size_width.fix(1.5)  # dimensionless
+        # No unfixed variable exist for crushing_direction_constraint,
+        # deactivate it to avoid assert_no_numerical_warning failure
+        m.fs.unit.crushing_direction_constraint[0].deactivate()
```
Daison2102, line 192: The main solve/diagnostic fixture disables the new stage-size constraints and deactivates `crushing_direction_constraint`, so the component tests do not exercise the main behavior this PR adds.

I guess this was done because: `_set_base_state` fixes the inlet and outlet particle-size variables, so `feed_p80` and `prod_p80` become expressions over fixed values. With the P80 inequality constraints active, `DiagnosticsToolbox.assert_no_numerical_warnings()` can report constraints with no unfixed variables. But that is exactly the interaction we need the tests to cover.

Can we add at least one solved/diagnostic test with the stage bounds and `crushing_direction_constraint` active? If the intended usage is to fix outlet particle size, the test should show that the active constraints behave cleanly for an in-range case. If the intended usage is for the solver to determine outlet size, the test should leave the relevant outlet size variable unfixed and add whatever equation/specification closes the model.

src/prommis/solid_handling/tests/test_crusher.py
Daison2102, line 209: Can we add `assert_units_consistent(model.fs.unit)` and a degrees-of-freedom assertion to this test class? The PR adds new constraints and unit conversions around P80, and those checks would make the test coverage closer to nearby PROMMIS unit tests.

src/prommis/solid_handling/tests/test_crusher.py
```diff
@@ -133,11 +240,98 @@ def test_solution(self, model):
...
+    assert len(warning_messages) == 1
+    assert "falls in the 'secondary' crushing range" in warning_messages[0]
+    assert ">= 10 cm" in warning_messages[0]
+    assert "falls in the 'secondary' crushing range" in caplog.text
+
+
+@pytest.mark.unit
+def test_check_applicability_warning_mentions_equipment(caplog):
+    m = _build_model(crusher_equipment="hammer_mill")
+    _set_base_state(m.fs.unit, feed_median_um=50000, prod_median_um=58)
```
Daison2102: `_set_base_state(m.fs.unit, feed_median_um=50000, prod_median_um=58)` fixes `particle_size_median` of the outlet to 58 µm, but the very next call `_fix_product_p80(m.fs.unit, 1 * pyunits.cm)` immediately re-fixes it to a different value. The `prod_median_um=58` has no effect and could mislead a reader -  can it be removed?

src/prommis/solid_handling/crusher.py
Lingyan90: removed

src/prommis/solid_handling/crusher.py
Lingyan90: changing only `_recommend_stage_from_product_p80_native()` could make the recommendation logic look inconsistent with `_product_p80_in_stage_range()`, because `check_applicability()` also uses `_product_p80_in_stage_range()` later to decide whether the selected stage is valid. In the current code, both recommendation and selected-stage checking share the same clipped interval logic.

This edge case is unlikely to occur in user-specified inputs because the eps gap is only 1e-6 cm, while all crusher stage are in cm scale. user input value of something ilke `9.9999991 cm` is very unlikely, more likely would put 10 cm instead. Therefore, it is not a major practical concern. However, the current recommendation logic still makes `in_modeled_range=False` mean either "below 0.5 cm" or "in a tiny numerical gap near a stage boundary." If we want the logic to be strictly unambiguous, recommendation/classification should avoid the eps-adjusted upper bounds. If we only care about practical user inputs, the current implementation is acceptable.

src/prommis/solid_handling/tests/test_crusher.py
Lingyan90: that's a good point. I'll add a test at the end of the test ifle in the new push.

src/prommis/solid_handling/tests/test_crusher.py
Lingyan90: added both under the `crushing_direction_constraint[0].deactivate` case and `crushing_direction_constraint.active` case.

src/prommis/solid_handling/tests/test_crusher.py
Lingyan90: removed

src/prommis/solid_handling/crusher.py
Lingyan90: changed it to `False` and also added `enforce_stage_size_limits=True` in test file during model build that specifically about stage bounds, e.g. `test_check_applicability_no_warning_for_in_range_secondary(caplog)` and a newly added test `test_solve_with_stage_bounds_and_crushing_direction_active()`

src/prommis/solid_handling/crusher.py
Lingyan90: There is no literature reference for defaulting an unspecified crusher to the secondary stage. That was a modeling choice used as a fallback when neither crusher_stage nor crusher_equipment is provided.

The stage product-P80 ranges and the equipment-to-stage mapping are based on the cited source in the module docstring. These are applicability guidance rather than universal hard validity limits.

src/prommis/solid_handling/crusher.py
Lingyan90: removed this line of code `selected_equipment = selected_equipment.strip().lower()`

src/prommis/solid_handling/crusher.py
Lingyan90: added `crusher_stage` `crusher_equipment` and `stage_selection_basis` as properties after `build()` in the new push.

src/prommis/solid_handling/crusher.py
Lingyan90: This was intended only as an initialization aid, not as a model specification. It does not fix the outlet particle-size median; it only sets an initial value when the variable is unfixed. The actual product size still needs to be specified by the user or determined by the model. I added comments above the code to clarify this.

src/prommis/solid_handling/crusher.py
Lingyan90: The intended usage supports both workflows.

If the outlet particle-size variables are free or partially free, the crushing-direction constraint acts as a hard algebraic constraint and the solver enforces `feed_p80 >= prod_p80`.

If the user fixes all particle-size variables, the same constraint becomes a feasibility check on the specified values. That is still logically valid, but it has no unfixed variables and can therefore appear as a zero-Jacobian row in IDAES numerical diagnostics. In that fully specified workflow, `check_applicability()` is the preferred way to provide user-facing warnings, while the hard constraint is mainly useful for solved/optimization cases.

I added a separate solve/diagnostic test with the outlet `particle_size_median` free to cover the active-constraint use case, and kept the fixed-size tests focused on post-hoc applicability checks.

src/prommis/solid_handling/crusher.py
Daison2102: Agreed that changing only the recommendation function would be inconsistent. The fix is to make the shared helper `_product_p80_in_stage_range()` gap-free instead, so the recommendation and `check_applicability()`'s selected-stage check stay in sync. Python supports strict `<`, so classification doesn't need the eps, express each stage as a half-open interval `[lower, next_lower)`. The eps stays only in `product_p80_upper_bound`, where it emulates the strict upper bound Pyomo can't. This makes `in_modeled_range=False` mean exactly "below 0.5 cm." Also agree that it is unlikely to get to this edge case, so fine to leave as it is. But the change is small to make it unambiguous. Look below. So, Up to you whether you want to apply it:

```python
def _product_p80_in_stage_range(stage_name, product_p80_native, native_units):
    lower, upper = _get_stage_bounds_native(stage_name, native_units)
    if lower is not None and product_p80_native < lower:
        return False
    if upper is not None and product_p80_native >= upper:
        return False
    return True
```

`_recommend_stage_from_product_p80_native()` and `product_p80_upper_bound` need no change.

Daison2102 (approved):
It looks good!

agarciadiego (approved):
LGTM

PR #264, opened 2026-06-02
Add flotation unit model files

Daison2102:
Bring in the flotation unit model, flowsheets, parameter data, and tests for the flotation-unit-model branch.

## Addresses Issue:

N/A

## Summary/Motivation:

Adds the Mountain Pass bastnaesite flotation modeling module to PrOMMiS. The module supports fixed-recovery and kinetic flotation-bank formulations, includes parameter data for reference scenarios, and provides tests for model structure, initialization, solve behavior, mass closure, and reference-result consistency.

- Add the `prommis.flotation` module with bastnaesite properties, flotation-bank unit model, initialization helpers, Mountain Pass flowsheets, and kinetic parameter data.
- Add flotation tests covering property-package behavior, unit-model behavior, fixed-recovery and kinetic flowsheets, parameter-data validation, mass closure, and reference scenario outputs.

- [x] Documentation
- [x] Tests
- [x] Diagnostic tests for models


src/prommis/flotation/mountain_pass_kinetic_closed_form_parameters.json
```diff
@@ -0,0 +1,161 @@
...
+  "bank_constants": {
+    "rho_water_kg_per_m3": 1000.0
+  },
+  "basis": {
+    "figure_2_units": "t/shift",
+    "shift_duration_h": 11.0,
+    "plant_throughput_t_per_day": 1700.0,
+    "plant_dry_feed_kg_per_h": 70833.33333333333,
+    "json_fresh_feed_total_kg_per_h": 789.5,
+    "scale_jsonbasis_to_plant_h": 89.71923158116951
```
Lingyan90: Is this level of decimal precision needed here? If not, it may be worth rounding this and similar values elsewhere in the PR for readability and consistency.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,222 @@
...
+# https://www.sciencedirect.com/science/article/abs/pii/S0892687525007083
+# kWh/t is equivalent to W*h/kg, the model's native units.
+BASTNAESITE_BOND_WORK_INDEX = 6.96
+SOLID_DENSITIES = {
+    "REO": 5000.0,
+    "CaO": 2710.0,
+    "BaO": 4480.0,
+    "SrO": 3960.0,
+    "inert_gangue": 2650.0,
+}
```
Lingyan90, line 45: Are these solid density values taken from a published reference or an empirical source? If so, could you please cite the source here? Also, could you clarify whether these are true/material densities, which should be independent of particle size, or bulk/apparent densities, which may vary with particle size and packing because of changes in void space?

src/prommis/flotation/mountain_pass_flotation_fixed_recovery_flowsheet.py
```diff
@@ -0,0 +1,603 @@
...
+            "Scenario data feed components do not match the property package."
+        )
+    for scenario in data["scenarios"].values():
+        for bank_recoveries in scenario["recoveries"].values():
+            if set(bank_recoveries) != set(COMPONENTS):
+                raise ValueError(
+                    "Scenario data recovery components do not match the property "
+                    "package."
+                )
+    return data
```
Lingyan90, line 112: The above code check that feed and recovery component from json sets match `COMPONENTS`, would be good to also check that the values are make sense. e.g. feed flows is above 0, initial stream flows >0, and recoveries are between 0 and 1.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+                # The tiny upper tolerance avoids diagnostics warning on the
+                # default fixed-at-one value while preserving the intended
+                # ultimate-recovery bound for estimation workflows.
+                bounds=(0, 1 + 1e-8),
+                units=units.dimensionless,
+                doc="Ultimate recoverable fraction fixed at 1.0 by default",
+            )
+            for time in self.flowsheet().time:
+                for component in component_list:
+                    # Fix R_inf for current implementation. IF rich recovery-time
```
Lingyan90:
```suggestion
                    # Fix R_inf for current implementation. If rich recovery-time
```

Lingyan90 (commented):
Overall, this looks good to me. I have a few comments and suggestions below.

sk-netl (approved):

ksbeattie:
@Daison2102, any news on this?

src/prommis/flotation/mountain_pass_kinetic_closed_form_parameters.json
Daison2102: All JSON values reduced to 5 decimals

src/prommis/flotation/bastnaesite_properties.py
Daison2102: Done

src/prommis/flotation/flotation_bank.py
Daison2102: Done

src/prommis/flotation/mountain_pass_flotation_fixed_recovery_flowsheet.py
Daison2102: Agree. Value validation added, with one correction to the bounds.

Daison2102:
> @Daison2102, any news on this?

Pending comments have been addressed. Now, it is ready.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,228 @@
...
+        obj.add_properties(
+            {
+                "flow_mass_comp": {"method": None},
+                "flow_mass": {"method": None},
+                "mass_frac_comp": {"method": "_mass_frac_comp"},
+            }
+        )
+        obj.define_custom_properties(
+            {
+                "flow_vol_solid": {"method": None},
```
dallan-keylogic: Why not just define `flow_vol_phase` over the single solid phase?

After reviewing the unit model, it seems like this is the flow rate of dry solids suspended in a slurry. It might be worth making that distinction in the property name.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,228 @@
...
+                self.flow_mass_comp[component] / self.params.rho_mass_comp[component]
+                for component in self.params.component_list
+            ),
+            doc="Volumetric flow of the solids phase",
+        )
+
+    def _mass_frac_comp(self):
+        # Note: this Expression is undefined for zero-flow streams (division by
+        # zero). It is used only for reporting; callers that may encounter
+        # zero-flow streams should guard before calling value().
```
dallan-keylogic, line 211: Then it would be better to make `mass_frac_comp` and `flow_mass` the state variables.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,228 @@
...
+        self.mass_frac_comp = Expression(
+            self.params.component_list,
+            rule=lambda b, j: b.flow_mass_comp[j] / b.flow_mass,
```
dallan-keylogic: Either define a rule using `def` or use decorator notation; using lambda functions is discouraged.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+``unit.default_initializer(strategy="staged").initialize(unit)``, or import and
+use either strategy class directly.
+
+Dynamic-flowsheet compatibility
+-------------------------------
+
+The ``fixed`` and ``kinetic`` modes are algebraic in time — every constraint
+is evaluated independently at each time point and the unit carries no
+accumulation state. They can be embedded in dynamic flowsheets without
+restriction; the state blocks propagate the flowsheet's dynamic context. Only
```
dallan-keylogic: My first instinct is that you shouldn't allow a `dynamic=True` option to be passed to a unit model without any holdup or accumulation variables. Steady state unit models can be included in dynamic flowsheets so long as the individual unit models use the `dynamic=False` option.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+
+    G_{t,i,j} = k_{t,j} M_{t,i,j}
+
+The total cell inventory is fixed by the geometric closure:
+
+.. math::
+
+    M_{t,i} = V_{cell}(1 - \epsilon_{air})\rho_{slurry}w_s
+
+and component holdups sum to ``M[t, i]``. A well-mixed algebraic closure is
```
dallan-keylogic: I'm not sure what "algebraic closure" means here.

src/prommis/flotation/initializer.py
```diff
@@ -0,0 +1,1342 @@
...
+                }
+            if abs(diag["lower_margin"] - 1.0) < MARGIN_NEAR_BOUNDARY:
+                summary["near_boundary_cells"] += 1
+    return summary
+
+
+def _safe_value(obj, default):
+    val = value(obj, exception=False)
+    if val is None or not math.isfinite(val):
+        return default
```
dallan-keylogic, line 411: This doesn't catch `NaN` values.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+                for (time, component), condata in constraint_obj.items():
+                    feed_nominal = _safe_value(
+                        model.properties_in[time].flow_mass_comp[component],
+                        default=1e-3,
+                    )
+                    self.set_constraint_scaling_factor(
+                        condata,
+                        1 / max(abs(feed_nominal), 1e-3),
+                        overwrite=overwrite,
+                    )
```
dallan-keylogic: These scaling factors should be calculated from the corresponding variable scaling factors, not by the present value of variables.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+        ConfigBlock(
+            implicit=True,
+            description="Arguments to use when constructing property blocks",
+        ),
+    )
+    CONFIG.declare(
+        "recovery_basis",
+        ConfigValue(
+            default="fixed",
+            domain=In(["fixed", "kinetic_closed_form", "kinetic_cell_balance"]),
```
dallan-keylogic: You should use a `StrEnum` from `idaes.core.util.misc`.

Lingyan90 (approved):
LGTM

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+                bounds=(1e-6, None),
+                units=units.m**3,
+                doc="Volume per flotation cell",
+            )
+            self.air_holdup = Var(
+                initialize=0.10,
+                bounds=(0, 0.5),
+                units=units.dimensionless,
+                doc="Volumetric air hold-up fraction in the cells",
+            )
```
dallan-keylogic: Consider renaming this `volume_frac_air` or something similar, to avoid confusion with the holdup terms used for dynamics.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+                # ultimate-recovery bound for estimation workflows.
+                bounds=(0, 1 + 1e-8),
+                units=units.dimensionless,
+                doc="Ultimate recoverable fraction fixed at 1.0 by default",
+            )
+            for time in self.flowsheet().time:
+                for component in component_list:
+                    # Fix R_inf for current implementation. If rich recovery-time
+                    # data are available, it can be estimated.
+                    self.R_inf[time, component].fix(1.0)
```
dallan-keylogic: We shouldn't fix variables when constructing a unit model, especially not when the variable is time-indexed. If the user creates a dynamic flowsheet and then adds additional time points as part of the discretization method (like is done with collocation), then we end up with degrees-of-freedom issues at the new time point because the variable is not fixed at those time points.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+                    / b.pulp_solids_mass_fraction
+                )
+
+            @self.Expression(self.flowsheet().time)
+            def flow_vol_water(b, t):
+                return b.flow_mass_water[t] / b.rho_water
+
+            @self.Expression(self.flowsheet().time)
+            def flow_vol_slurry(b, t):
+                return b.properties_in[t].flow_vol_solid + b.flow_vol_water[t]
```
dallan-keylogic: How difficult would it be to create the slurry-aware property package now? I'd suggest making phase-component mass fractions and phase mass flow rates as the state variables.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+
+            @self.Expression(self.flowsheet().time)
+            def tau(b, t):
+                return b.effective_volume / b.flow_vol_slurry[t]
```
dallan-keylogic, line 702: Give this a more descriptive name, like `residence_time`.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+            @self.Constraint(self.flowsheet().time, component_list)
+            def tails_split_eq(b, t, j):
+                return (
+                    b.properties_tails[t].flow_mass_comp[j]
+                    == b.properties_in[t].flow_mass_comp[j]
+                    - b.properties_concentrate[t].flow_mass_comp[j]
+                )
+
+        @self.Expression(self.flowsheet().time)
+        def solid_mass_pull(b, t):
```
dallan-keylogic, line 740: Is this supposed to be named "solid_mass_**pulp**"?

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+            bounds=(1e-6, None),
+            units=units.m**3,
+            doc="Volume per flotation cell",
+        )
+        self.air_holdup = Var(
+            initialize=0.10,
+            bounds=(0, 0.5),
+            units=units.dimensionless,
+            doc="Volumetric air hold-up fraction in the cells",
+        )
```
dallan-keylogic: Same thing here with "air_holdup"->"volume_frac_air"

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+        self.cell_total_solid_holdup = Var(
+            self.flowsheet().time,
+            self.cells,
+            initialize=1.0,
+            bounds=(0, None),
+            units=units.kg,
+            doc="Total dry-solid holdup in each cell",
+        )
+        self.cell_solid_holdup = Var(
```
dallan-keylogic, line 807: Rename to "solid_holdup_cell_comp"

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+            component_names,
+            initialize=1e-3,
+            bounds=(0, None),
+            units=units.kg,
+            doc="Component dry-solid holdup in each cell",
+        )
+        self.cell_pulp_out_flow = Var(
+            self.flowsheet().time,
+            self.cells,
+            component_names,
```
dallan-keylogic: Rename to "pulp_out_flow_cell_comp"

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+            initialize=1e-3,
+            bounds=(0, None),
+            units=units.kg / units.hour,
+            doc="Component dry-solid pulp outlet flow from each cell",
+        )
+        self.cell_float_flow = Var(
```
dallan-keylogic: Should be renamed "float_flow_cel_comp"

src/prommis/flotation/initializer.py
```diff
@@ -0,0 +1,1342 @@
...
+                body_val = value(con.body, exception=False)
+                upper_val = (
+                    value(con.upper, exception=False) if con.upper is not None else 0.0
+                )
+            except (ValueError, TypeError):
+                continue
+            if body_val is None or upper_val is None:
+                continue
+            if not (math.isfinite(body_val) and math.isfinite(upper_val)):
+                continue
```
dallan-keylogic: What is the logic behind the `continue` statements here? I understand skipping the constraint if it isn't active, but why not raise an exception if a constraint doesn't exist? Why are we wrapping `body_val` and `upper_val` in a try-except block if we're explicitly telling `value` not to raise an exception? Why skip over a constraint if its body does not have a value? Why skip over a constraint if either its body or upper bound has a value of infinity?

src/prommis/flotation/initializer.py
```diff
@@ -0,0 +1,1342 @@
...
+    F_in_cell_raw,
+    M_total,
+    k_hour,
+    cell_index,
+    context,
+    *,
+    collect_diagnostics,
+):
+    """Solve one cell in the fixed-inventory kinetic cascade."""
+    from scipy.optimize import brentq
```
dallan-keylogic: Why are you deferring the import here?

dallan-keylogic (changes requested):
Here is some preliminary feedback. This is a very big PR with a lot of content, so I'll need some extra time to review it all.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,228 @@
...
+# dry-solid mass flow to volumetric flow. Source: Handbook of Mineralogy
+# (Anthony, Bideaux, Bladh & Nichols, Mineral Data Publishing, 2001-2005).
+SOLID_DENSITIES = {
+    "REO": 5000.0,
+    "CaO": 2710.0,
+    "BaO": 4480.0,
+    "SrO": 3960.0,
+    "inert_gangue": 2650.0,
+}
+REFERENCE_SCALING_FACTORS = {component: 1.0 for component in COMPONENTS}
```
dallan-keylogic, line 46: Why define this as a dictionary separate from `DEFAULT_SCALING_FACTORS`.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def variable_scaling_routine(
+        self, model, overwrite: bool = False, submodel_scalers: dict = None
+    ):
+        for component in model.params.component_list:
+            self.set_variable_scaling_factor(
+                model.flow_mass_comp[component],
+                REFERENCE_SCALING_FACTORS[component],
+                overwrite=overwrite,
+            )
```
dallan-keylogic, line 67: Since we've got entries for these variables in the `default_scaling_factors` dictionary, these should be scaled by default.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+    ``recovery_basis="kinetic_closed_form"`` and ``"kinetic_cell_balance"``. In
+    fixed mode it is accepted but does not add kinetic variables, expressions,
+    or constraints.
+
+Degrees of Freedom
+------------------
+
+For each time point and component, fixed mode has one degree of freedom in the
+local recovery variable, ``recovery[t, j]``. Flowsheets typically fix these
+recoveries from plant data or add process constraints that determine them.
```
dallan-keylogic, line 64:
```suggestion
When using the ``"fixed"`` recovery basis, there is one degree of freedom 
for each time point and component : the local recovery variable, ``recovery[t, j]``. 
Users are required to fix these recoveries from plant data or add process 
constraints that determine them.
```

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+
+Closed-form kinetic mode (``"kinetic_closed_form"``) is a prediction form.
+Users fix ``k_cf[t, j]`` and leave ``recovery[t, j]`` free. They also fix
+``cell_volume``, ``air_holdup``, and ``pulp_solids_mass_fraction`` for each
+bank.
```
dallan-keylogic, line 69:
```suggestion
When using the ``"kinetic_closed_form"``) recovery basis, users should fix
``k_cf[t, j]`` and leave ``recovery[t, j]`` free. They should also fix ``cell_volume``,
``air_holdup``, and ``pulp_solids_mass_fraction`` for each bank.
```

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+
+Kinetic-cell-balance mode (``"kinetic_cell_balance"``) is a prediction form.
+Users fix ``cell_volume``, ``air_holdup``, ``pulp_solids_mass_fraction``,
+``rho_solid``, ``rho_water``, and ``k_cb[t, j]``; the cell balances then
+determine cell holdups, outlet flows, concentrate and tails flows, and
+``recovery[t, j]``. This mode is steady-state only and rejects ``dynamic=True``.
```
dallan-keylogic:
```suggestion
When using the ``"kinetic_cell_balance"`` recovery basis,
users fix ``cell_volume``, ``air_holdup``, ``pulp_solids_mass_fraction``,
``rho_solid``, ``rho_water``, and ``k_cb[t, j]``; the cell balances then
determine cell holdups, outlet flows, concentrate and tails flows, and
``recovery[t, j]``. This mode is steady-state only and rejects ``dynamic=True``.
```

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+
+The bank's ``default_initializer`` is
+``FlotationBankKineticCellBalanceInitializer``, a dispatcher that selects
+between two interchangeable strategies via its ``strategy`` config option:
```
dallan-keylogic, line 79:
```suggestion
between two strategies via its ``strategy`` config option:
```

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,951 @@
...
+  ``cell_cascade_forward`` to compute a constraint-satisfying cascade in pure
+  Python and seed it directly (no solver call).
+* ``strategy="staged"`` runs
+  ``FlotationBankKineticCellBalanceStagedInitializer``, which applies the
+  standard IDAES staged fix / deactivate / solve / unfix / reactivate pattern,
+  solving the cell cascade through the Pyomo equations themselves.
+
+Select the staged strategy with
+``unit.default_initializer(strategy="staged").initialize(unit)``, or import and
+use either strategy class directly.
```
dallan-keylogic, line 92: I think having the parent class here is redundant. Just provide both strategies as independent `Initializer` objects.

src/prommis/flotation/initializer.py
```diff
@@ -0,0 +1,1342 @@
...
+        offset = REL_BRACKET_EPS * max(F_total - F_zero, F_zero)
+    else:
+        lower_boundary = 0.0
+        offset = REL_BRACKET_EPS * F_total
+
+    for _ in range(BRACKET_REFINE_MAX_ITER):
+        T_low = lower_boundary + offset
+        if T_low > 0 and g(T_low) > 0:
+            return T_low, T_high
+        offset *= 0.5
```
dallan-keylogic, line 113: I don't understand what this algorithm is accomplishing. What I *think* this function is doing is to create a lower bound for a bracketing method that has the opposite sign (positive) as the upper bound (negative). However, I do not get what's going on when adding `offset`. Its value is on the order of `1e-9` because it's being multiplied by `REL_BRACKET_EPS`. The lower bound `T_low` keeps getting `offset` added to it, and `offset` keeps getting smaller. The largest bound that `T_low` can take is `lower_boundary + 2*offset`. Do we have any reason to think an acceptable point will lie in this domain?

src/prommis/flotation/initializer.py
```diff
@@ -0,0 +1,1342 @@
...
+
+    F_in_cell = {
+        j: (F_in_cell_raw[j] if F_in_cell_raw[j] > FEED_FLOOR_KG_PER_H else 0.0)
+        for j in F_in_cell_raw
+    }
+    F_total = sum(F_in_cell.values())
+
+    if F_total <= FEED_FLOOR_KG_PER_H:
+        # The depleted_cell mode has no meaningful "margin deficit" (lower_margin
+        # is undefined when there is no cell feed at all). We report
```
dallan-keylogic, line 147: What is a "margin deficit"?

src/prommis/flotation/bastnaesite_properties.py
Daison2102: Addressed. This keeps the dry-solid meaning explicit through the `solid` phase index.

src/prommis/flotation/bastnaesite_properties.py
Daison2102: I am keeping `flow_mass_comp` as the state variable. For this dry-solids package, component mass flows avoid an extra sum-to-one constraint and avoid undefined mass fractions at zero total flow. A total-flow/composition state set may make sense for a future slurry-aware package, but I do not think it is the right change here.

src/prommis/flotation/bastnaesite_properties.py
Daison2102: Addressed. I replaced the lambda with a named rule function for `mass_frac_comp`.

src/prommis/flotation/flotation_bank.py
Daison2102: Addressed. `FlotationBank` now restricts `dynamic=False` and `has_holdup=False` for all recovery bases. The unit remains usable inside a dynamic flowsheet as a steady-state unit model.

src/prommis/flotation/flotation_bank.py
Daison2102: I kept the term `algebraic closure` but clarified that it is a steady-state composition relation linking cell solid holdup to pulp outlet composition, not a dynamic accumulation equation.

src/prommis/flotation/initializer.py
Daison2102: I did not change this. `_safe_value` already uses `math.isfinite`, which returns `False` for numeric NaN values, so numeric NaNs already fall back to the default.

src/prommis/flotation/flotation_bank.py
Daison2102: Addressed. I removed the nominal-value fallback from constraint scaling. The scaler now requires variable scaling factors for the variables used in the constraints, combines those factors for product terms, and raises `ConfigurationError` if a required factor is missing.

src/prommis/flotation/flotation_bank.py
Daison2102: Done.

src/prommis/flotation/flotation_bank.py
Daison2102: Renamed. The calibration JSON key remains `air_holdup` because that is source-data terminology, but the unit-model API now uses the clearer IDAES-facing name.

src/prommis/flotation/flotation_bank.py
Daison2102: Addressed by converting `R_inf` from a fixed Var to a mutable `Param`. This keeps the current examples at zero DOF without fixing a time-indexed variable during unit construction. If future data support estimating `R_inf`, we can promote it back to a Var.

src/prommis/flotation/flotation_bank.py
Daison2102: I am not adding a slurry-aware property package in this PR. The current model is scoped as a dry-solids flotation model with bank-local inferred water quantities for residence-time and inventory estimates only. This is mainly because the paper used to validate this unit model does not provide enough data for this purpose, so it does not make sense to develop it at this point. A water-conserving slurry property package should be treated as a separate design task if data become available for validation.

src/prommis/flotation/flotation_bank.py
Daison2102: Done.

src/prommis/flotation/flotation_bank.py
Daison2102: I am keeping `solid_mass_pull`. In flotation, mass pull is the intended term for the fraction of feed mass reporting to concentrate. Renaming it to `pulp` would be misleading because this model is not tracking slurry pulp as a conserved stream phase.

src/prommis/flotation/flotation_bank.py
Daison2102: Addressed. I also renamed the kinetic-cell-balance mode attribute to `volume_frac_air` for consistency across both kinetic formulations.

src/prommis/flotation/flotation_bank.py
Daison2102: I am keeping `cell_solid_holdup`. My naming choice was to put `cell_` first for per-cell quantities, then name the physical quantity. I am not adding `_comp` here because `flow_mass_comp` is a specific IDAES convention for component mass-flow variables, not a required suffix for every component-indexed quantity. The component index on `cell_solid_holdup[t, cell, comp]` already carries that information.

src/prommis/flotation/flotation_bank.py
Daison2102: Partially addressed. I renamed this to `cell_pulp_out_flow_mass_comp`, not `pulp_out_flow_cell_comp`. I kept `cell_` first because this is a per-cell submodel quantity, and I used the IDAES-style `flow_mass_comp` suffix because the variable is a component mass flow.

src/prommis/flotation/flotation_bank.py
Daison2102: Partially addressed. I renamed this to `cell_flotation_flow_mass_comp`, not `float_flow_cel_comp`. This uses the process term `flotation`, keeps the per-cell context first, and uses the `flow_mass_comp` convention for component mass-flow variables.

src/prommis/flotation/initializer.py
Daison2102: Partially addressed. The residual checker still skips inactive constraints because the staged initializer intentionally deactivates outlet constraints during the cell-balance sub-block solve. It now raises `ConfigurationError` for missing expected constraints, unevaluable active constraints, or non-finite residuals.

src/prommis/flotation/initializer.py
Daison2102: Addressed. I moved `brentq` to module scope. The original deferred import was meant to avoid loading `scipy.optimize` unnecessarily, but this initializer already imports IDAES initialization utilities that load `scipy.optimize`, so the local import did not avoid the cost or make SciPy optional.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,236 @@
...
+        if model.params.config.has_particle_size_distribution:
+            self.set_variable_scaling_factor(
+                model.particle_size_median,
+                1e-5,
+                overwrite=overwrite,
+            )
```
dallan-keylogic, line 73: This should also be added to the default scaling factors dictionary and scaled by default.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,1009 @@
...
+
+    G_{t,i,j} = k_{t,j} M_{t,i,j}
+
+The total cell inventory is fixed by the geometric closure:
+
+.. math::
+
+    M_{t,i} = V_{cell}(1 - \epsilon_{air})\rho_{slurry}w_s
+
+and component holdups sum to ``M[t, i]``. A well-mixed algebraic closure, meaning
```
dallan-keylogic, line 241: The term "algebraic closure" is guaranteed to be confusing. It has [several](https://en.wikipedia.org/wiki/Algebraic_closure) [meanings](https://en.wikipedia.org/wiki/Algebraic_closure_(convex_analysis)), neither of which is being used here.

src/prommis/flotation/mountain_pass_flotation_fixed_recovery_flowsheet.py
```diff
@@ -0,0 +1,614 @@
...
+        data = json.load(file)
+    fresh_feed = data["feed"]["fresh_feed"]
+    if set(fresh_feed) != set(COMPONENTS):
+        raise ValueError(
+            "Scenario data feed components do not match the property package."
+        )
+    if any(flow < 0 for flow in fresh_feed.values()):
+        raise ValueError("Scenario data fresh feed flows must be non-negative.")
+    if stream_total(fresh_feed) <= 0:
+        raise ValueError("Scenario data fresh feed total flow must be positive.")
```
dallan-keylogic, line 97:
```suggestion
```
If all of the component flows are strictly positive, then their sum must be too.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,1009 @@
...
+                    nominal = _safe_value(var, default=feed_nominal)
+                    self.set_variable_scaling_factor(
+                        var, 1 / max(abs(nominal), 1e-12), overwrite=overwrite
+                    )
+
+    def _required_scaling_factor(self, var):
+        scaling_factor = self.get_scaling_factor(var)
+        if scaling_factor is not None and scaling_factor > 0:
+            return scaling_factor
+        raise ConfigurationError(f"Missing scaling factor for {var.name}.")
```
dallan-keylogic, line 402: Should we add a flag to `get_scaling_factor` to raise an exception if none exists?

src/prommis/flotation/tests/test_mountain_pass_flotation_kinetic_closed_form_flowsheet.py
```diff
@@ -0,0 +1,341 @@
...
+    with caplog.at_level(logging.INFO, logger=__name__):
+        for mixer_name, residual in diagnostics.items():
+            _log.info(
+                "Apparent water non-conservation diagnostic for %s: %.6g kg/h",
+                mixer_name,
+                residual,
+            )
+
+    assert any(abs(residual) > 1e-6 for residual in diagnostics.values())
+    assert "Apparent water non-conservation diagnostic" in caplog.text
```
dallan-keylogic, line 243: Is this diagnostic a method on the floatation bank or Mountain Pass flowsheet? If so, that method should be tested here. If not, what is being tested here?

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,1009 @@
...
+
+    def _scale_constraint_by_required_component(
+        self, target_constraint, scaling_component, overwrite
+    ):
+        self.set_constraint_scaling_factor(
+            target_constraint,
+            self._required_scaling_factor(scaling_component),
+            overwrite=overwrite,
+        )
```
dallan-keylogic, line 411: This could also be added as an option to `scale_constraint_by_component`.

src/prommis/flotation/flotation_bank.py
```diff
@@ -0,0 +1,1009 @@
...
+    """
+
+    default_scaler = FlotationBankScaler
+    default_initializer = BlockTriangularizationInitializer
+
+    CONFIG = UnitModelBlockData.CONFIG()
+    CONFIG.get("dynamic")._domain = In([False])
+    CONFIG.get("dynamic")._default = False
+    CONFIG.get("has_holdup")._domain = In([False])
+    CONFIG.get("has_holdup")._default = False
```
dallan-keylogic, line 553: The last time I discussed this with the Pyomo team, they said that we should use `del CONFIG.dynamic` and then add it as a new config value with the desired domain and default.

src/prommis/flotation/initializer.py
```diff
@@ -0,0 +1,1358 @@
...
+
+def _solve_cell(
+    F_in_cell_raw,
+    M_total,
+    k_hour,
+    cell_index,
+    context,
+    *,
+    collect_diagnostics,
+):
```
dallan-keylogic, line 137: Could you please add some additional comments to this method to document what each step is doing?

src/prommis/flotation/initializer.py
```diff
@@ -0,0 +1,1358 @@
...
+            init_log.info_high("Initialization Complete.")
+
+            return None
+        finally:
+            # Roll back any state changes that did not get cleared in the
+            # success path (e.g., when stage 3 raised).
+            for var in temp_fixed_outlet_vars:
+                var.unfix()
+            for cdata in deactivated_outlet_cons:
+                cdata.activate()
```
dallan-keylogic, line 1287: Why do we need this `finally` block when the initializer object already rolls back variable fixedness and constraint activity upon initialization failing?

src/prommis/flotation/mountain_pass_flotation_fixed_recovery_flowsheet.py
```diff
@@ -0,0 +1,614 @@
...
+        }
+    )
+    return streams
+
+
+def stream_port_map(model):
+    fs = model.fs
+    # StateJunction inlet/outlet ports intentionally alias the same state vars.
+    # Listing both ports keeps the named stream map complete for initialization
+    # and reporting without adding pass-through constraints.
```
dallan-keylogic, line 336: You should be able to obtain this map by iterating over `m.fs.component_data_objects(ctype=Arc, descend_into=False)`, then looking at `arc.source` and `arc.destination`.

src/prommis/flotation/mountain_pass_flotation_fixed_recovery_flowsheet.py
```diff
@@ -0,0 +1,614 @@
...
+        "cleaner1_concentrate": [
+            fs.cleaner1.concentrate,
+            fs.cleaner2_mixer.cleaner1_concentrate,
+        ],
+        "cleaner1_tails": [fs.cleaner1.tails, fs.scavenger.inlet],
+        "scavenger_concentrate": [
+            fs.scavenger.concentrate,
+            fs.regrind.inlet,
+            fs.regrind.outlet,
+            fs.rougher_mixer.regrind_recycle,
```
dallan-keylogic, line 355: Why does this arc have four parts? Is it because `m.fs.regrind` is a state junction?

src/prommis/flotation/mountain_pass_flotation_fixed_recovery_flowsheet.py
```diff
@@ -0,0 +1,614 @@
...
+        raise ValueError(
+            f"Unknown initialization method {initialization_method!r}. "
+            f"Expected one of {INITIALIZATION_METHODS}."
+        )
+    model.fs.initialization_method = initialization_method
+
+
+def solve_model(model, solver=None, tee=False):
+    """Solve the flowsheet."""
+    opt = get_solver("ipopt") if solver is None else solver
```
dallan-keylogic, line 429: Any reason for using `ipopt` instead of `ipopt_v2`?

src/prommis/flotation/mountain_pass_flotation_kinetic_cell_balance_flowsheet.py
```diff
@@ -0,0 +1,503 @@
...
+
+TABLE1_PRODUCT_FIT_SCENARIO = recovery_flowsheet.TABLE1_PRODUCT_FIT_SCENARIO
+UNIT_MODEL_DEFAULT_INITIALIZATION = recovery_flowsheet.UNIT_MODEL_DEFAULT_INITIALIZATION
+DETERMINISTIC_STREAM_INITIALIZATION = (
+    recovery_flowsheet.DETERMINISTIC_STREAM_INITIALIZATION
+)
+INITIALIZATION_METHODS = recovery_flowsheet.INITIALIZATION_METHODS
+BANK_NAMES = recovery_flowsheet.BANK_NAMES
+FINAL_SOLVER_OPTIONS = {
+    "nlp_scaling_method": "user-scaling",
```
dallan-keylogic, line 78: You should be able to use the `.nl` writer-time scaling if you use `ipopt_v2`.

dallan-keylogic (approved):
Some more comments. After these are addressed this should be good to go.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,236 @@
...
+                "flow_mass_comp": {"method": None},
+                "flow_mass": {"method": None},
+                "flow_vol_phase": {"method": None},
+                "mass_frac_comp": {"method": "_mass_frac_comp"},
+            }
+        )
+        obj.define_custom_properties(
+            {
+                "particle_size_median": {"method": None},
+                "particle_size_width": {"method": None},
```
dallan-keylogic, line 134: One more thing---give these properties a method that raises a `ConfigurationError` telling the user to set `has_particle_size_distribution=True` if they want to use them.

src/prommis/flotation/bastnaesite_properties.py
```diff
@@ -0,0 +1,236 @@
...
+    )
+
+    def build(self):
+        super().build()
+
+        self.solid = Phase()
+        for component in COMPONENTS:
+            setattr(self, component, Component())
+
+        self.rho_mass_comp = Param(
```
adam-a-a, line 102:
```suggestion
        self.dens_mass_comp = Param(
```
IDAES naming convention

PR #268, opened 2026-06-15
Update nanofilteration values from watertap nf_brine update

sufikaur:
## Addresses Issue: 
Issue found upon pointing to main branches of watertap-org and idaes-pse

## Summary/Motivation:
A prior update to the dielectric constant of water used in MCAS affected the NF (NF brine) model’s Born energy contribution. Because the membrane pore dielectric constant is used via the unit model for the Gibbs energy calculation, the expected outputs in `nf_brine.py` shifted slightly, causing the existing tests to fail.

- update test values 

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


adam-a-a:
@dallan-keylogic I think this is what might be breaking tests: https://github.com/watertap-org/watertap/commit/00b5344671a1a79cf933df12d41ef8948dea280b#diff-4b5fe003ad0128ad3cc082b8d7c4d64296d8d634b04dcc00538592a59628866e

Specifically, I think the updated value for dielectric constant of water in MCAS is causing the tests to fail for nf_brine.py. I think it is reasonable and expected that pressure values would change a bit, for example. So my suggestion for anyone who is working on this would be to just update the test values (as long as none are brow-raising).

dallan-keylogic:
> @dallan-keylogic I think this is what might be breaking tests: [watertap-org/watertap@00b5344#diff-4b5fe003ad0128ad3cc082b8d7c4d64296d8d634b04dcc00538592a59628866e](https://github.com/watertap-org/watertap/commit/00b5344671a1a79cf933df12d41ef8948dea280b#diff-4b5fe003ad0128ad3cc082b8d7c4d64296d8d634b04dcc00538592a59628866e)
> 
> Specifically, I think the updated value for dielectric constant of water in MCAS is causing the tests to fail for nf_brine.py. I think it is reasonable and expected that pressure values would change a bit, for example. So my suggestion for anyone who is working on this would be to just update the test values (as long as none are brow-raising).

Thank you for pointing this out. I just traced back the test failure to that Davies PR. I'll change it back to its original value and see if the tests pass. So long as we understand why the tests are now failing, we can update the values.

dallan-keylogic:
I changed the dielectric constant back and the tests passed. It looks like, even if an ideal EoS is used, the dielectric constant is still used in calculating the Born energy as part of the NF model. It's a bit weird that the unit model is involved in property calculations, but it would  be hard to take into account the Gibbs energy difference between the bulk liquid and the pores without it.

Since we know what caused the difference in values, we can just update the test with the new values.

adam-a-a:
> I changed the dielectric constant back and the tests passed. It looks like, even if an ideal EoS is used, the dielectric constant is still used in calculating the Born energy as part of the NF model. It's a bit weird that the unit model is involved in property calculations, but it would be hard to take into account the Gibbs energy difference between the bulk liquid and the pores without it.
> 
> Since we know what caused the difference in values, we can just update the test with the new values.

Right- it's the dielectric constant of the membrane pore that requires the Gibbs energy calc to be in the unit model.

bpaul4 (approved):

PR #259, opened 2026-05-22
Improve UKy Recovery

MarcusHolly:
## Addresses Issue: 
#152

## Summary/Motivation:
Optimize the conditions of the UKy flowsheet to increase recovery from 1.4% to 22.2%

## Changes proposed in this PR:
- Updates the optimization function
- Updates the base operating conditions
- Updates UKy testing (WIP)

## Reviewer's checklist / merge requirements:
- [ ] The head branch (i.e. the "source" of the changes) is not the `main` branch on the PR author's fork
- [ ] Documentation
- [ ] Tests
- [ ] Diagnostic tests for models

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.md file
   at the top level of this directory.

2. I represent I am authorized to make the contributions and grant the license. If my employer has
   rights to intellectual property that includes these contributions, I represent that I have
   received permission to make contributions and grant the required license on behalf of that
   employer.

PR #291, opened 2026-09-10
Uky costing tutorial update

Lingyan90:
## Addresses Issue: 
updates the costing tutorial according to the new QGESS method.

## Summary/Motivation:
This update is to satisfy for a deliverable.
The REE recovery rate are been significantly improved in [PR#259](https://github.com/prommis/prommis/pull/259). This result in process and costing related value changes in this tutorial.

- Update the QGESS costing method from `ree_costing.py`.
- Added `get_total_BEC` and `display_flowsheet_cost` function to print out results similar to old tutorial version. These two function could be removed once `ree_costing.py` or 'QGESS.py` has these two function.

- [ ] Documentation
- [ ] Tests

### Legal Acknowledgement

   employer.


docs/tutorials/costing_uky_flowsheet-solution.ipynb
```diff
@@ -44,7 +44,19 @@
...
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "WARNING: DEPRECATED: the 'DiagnosticsToolbox' class has been moved to\n",
+      "'idaes.co re.util.diagnostics_tools.diagnostics_toolbox.DiagnosticsToolbox'.\n",
+      "Please update your import.  (deprecated in 2.12.0) (called from <frozen\n",
+      "importlib._bootstrap>:488)\n",
```
bpaul4, line 55: This import should be updated.

docs/tutorials/costing_uky_flowsheet-solution.ipynb
```diff
@@ -44,7 +44,19 @@
...
+      "2026-09-10 14:25:37 [INFO] idaes.prommis.costing.ree_costing_dictionaries: Custom REE plant currency units (USD_2025, USD_UKy_2019) already appear in Pyomo unit registry. Assuming repeated call of register_ree_currency_units.\n"
```
bpaul4, line 56: I'm not sure why it says "idaes.prommis", that location doesn't exist.

Regardless, this should not be thrown here, the currency units should not be registered on import.

docs/tutorials/costing_uky_flowsheet-solution.ipynb
```diff
@@ -1874,111 +1737,153 @@
...
+      "Number of equality constraint Jacobian evaluations   = 101\n",
       "Number of inequality constraint Jacobian evaluations = 0\n",
-      "Number of Lagrangian Hessian evaluations             = 60\n",
-      "Total CPU secs in IPOPT (w/o function evaluations)   =      0.257\n",
-      "Total CPU secs in NLP function evaluations           =      0.022\n",
+      "Number of Lagrangian Hessian evaluations             = 99\n",
+      "Total CPU secs in IPOPT (w/o function evaluations)   =      0.412\n",
+      "Total CPU secs in NLP function evaluations           =      0.033\n",
       "\n",
       "EXIT: Optimal Solution Found.\n"
```
bpaul4, line 1888: Wondering if scaling would help the solver out here, those regularization lines don't look great. Not critical though.

docs/tutorials/costing_uky_flowsheet-solution.ipynb
```diff
@@ -2085,42 +1971,195 @@
...
-    "QGESSCostingData.display_bare_erected_costs(m.fs.costing)"
+    "REECostingData.display_bare_erected_costs(m.fs.costing)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2069f643-2c55-4426-bd95-de57279d3ebe",
+   "metadata": {},
+   "source": [
+    "Define a flowsheet costing display method here. This could be changed once `ree_costing.py` or `QGESS.py` has this `display_flowsheet_cost()` method."
```
bpaul4, line 2014: Let's put this method in ree_costing.py for now, like we do with the report() method. That way we only have to undo this in that file later.

docs/tutorials/costing_uky_flowsheet-solution.ipynb
Lingyan90: The import in this tutorial is up to date, the file that gives warning is coming from other files that's are not updated yet, for this warning particularly, the `uky_flowsheet.py` was still using deprecated import. 

updated the `DiagnosticsToolbox` import directory in `uky_flowsheet.py`.

docs/tutorials/costing_uky_flowsheet-solution.ipynb
Lingyan90: the "idaes.prommis" is a logger name, not necessary means a directory for a file.

For the currency units that was registered on import, this registration excecutes wihle the class is defined, and therefore while the module is imported. 

Move the `register_ree_currency_units()` to under `build_global_params` can resolve this repeated call. 

But since `uky_flowsheet.py` is still using `ree_plant_capcost.py` instead of `ree_costing.py`, in the following cell 3, where `REECosting` is been used, it'll register the currency again. while Importing `prommis.uky.uky_flowsheet` also imports the legacy [prommis.uky.costing.ree_plant_capcost](https://github.com/prommis/prommis/blob/main/src/prommis/uky/uky_flowsheet.py), which registers the same units. This should resolve itself once the `uky_flowsheet.py` is updated with up-to-date costing file.

docs/tutorials/costing_uky_flowsheet-solution.ipynb
Lingyan90: do we want this type of flowsheet costing display in `ree_costing.py` only or more generally, in `QGESS.py` so other technology/flowsheet application could also use it.

bpaul4: Yes, having general report methods in `QGESS.py` would be more accessible for other projects, but that won't happen until the next IDAES release so for now let's put it here so PrOMMiS has access to it.

docs/tutorials/costing_uky_flowsheet-solution.ipynb
bpaul4: That makes sense, thanks for explaining. I'm okay leaving it for now then.

MarcusHolly (approved):
LGTM

bpaul4 (approved):

